PR #25, opened 2016-01-06
Fix #17: gb18030 cannot encode U+E5E5

### LINKED ISSUE #17: "gb18030 ranges" have problematic definitions
It seems that for some code points, gb18030 ranges doesn't
have a round-trip mapping.  Take the code point U+8000 for example.
  
When we apply the the "index gb18030 ranges pointer" we get:

    32768 ---> 18962, 0x4DAF --> 18962 + 32768 - 19887 --> 31843

but when we apply the "index gb18030 ranges code point" from
31843 we get:

    31843 ---> 19043, 0x9FA6 -->  40870 + 31843 - 19043 --> 53670

and that differs from our original 32768.  I think the reason is that each range is poorly defined; it's not clear where each range starts and ends in "index-gb18030-ranges.txt".

annevk:
Because of deployed content index gb18030 maps 0xA3 0xA0 to U+3000
rather than U+E5E5 when decoding. Therefore encoding U+E5E5 cannot work
either.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-encoding.svg" width="100"></a></p>
 <h1>Encoding</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-16-december-2015">Living Standard — Last Updated 16 December 2015</h2>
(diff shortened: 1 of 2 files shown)


inexorabletash:
lgtm

peteroupc:
Is it still true that Web sites using GB18030 have issues with this particular code point (decoding to U+3000 rather than U+E5E5)?  The Bugzilla issue mentioned in #17 is from 2002.

annevk:
@peteroupc I don't know. But it's not worth the churn of changing it at this point either I think.

PR #34, opened 2016-02-10
Fix #15: prevent ISO-2022-JP encoder attack

### LINKED ISSUE #15: iso-2022-jp encoder XSS risks
https://encoding.spec.whatwg.org/#iso-2022-jp-encoder

> 13\.2\.2 iso\-2022\-jp encoder

Suppose the input unicode string is A, B, ESC, $, B, 1, 2.

In this case, resulting encoded bytes will be A, B, ESC, $, B, 1, 2, according to the current spec.

The byte sequence can be, when decoded again, a unicode string A, B, 憶, which is apparently different from the original input unicode string.

The point I am making here is that, when encoding, ESC chars in the input unicode string should be specially handled (removed or replaced with something else). Or am I misunderstanding something?

Thanks,
Takeshi

annevk:


Overview.src.html
```diff
@@ -2214,6 +2214,15 @@ <h2 class="no-num no-toc">Table of Contents</h2>
  <span>ISO-2022-JP encoder state</span> is
  <span title="iso-2022-jp encoder ASCII">ASCII</span>, return <span>finished</span>.
 
+ <li>
+  <p>If <span>ISO-2022-JP encoder state</span> is either
+  <span title="iso-2022-jp encoder ASCII">ASCII</span> or
+  <span title="iso-2022-jp encoder Roman">Roman</span> and <var>code point</var> is U+000E, U+000F,
```
annevk: Add a comma after Roman.

PR #35, opened 2016-02-10
Fix #29: introduce output encoding for URL and HTML forms

### LINKED ISSUE #29: Introduce "output encoding"
@hsivonen pointed out that when HTML ends up using e.g., the "replacement" encoding to encode something, we would want to label that as `utf-8`, not `replacement`. Therefore, we should introduce a new concept called "output encoding" that is the encoding itself except for `replacement` (and soon hopefully utf-16le/be).

Then we remove replacement's encoder. Make sure the encode hook uses the output encoding of an encoding. And make sure HTML and URL use the encode hook correctly.

annevk:
Also remove the replacement encoder which would be weird to use anyway
as the encoding label would be “replacement” rather than “utf-8”.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -40,7 +40,8 @@
  <li><a href="#encodings"><span class="secno">5 </span>Encodings</a>
   <ol>
    <li><a href="#encoders-and-decoders"><span class="secno">5.1 </span>Encoders and decoders</a></li>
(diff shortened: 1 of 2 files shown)


annevk:
See http://logs.glob.uno/?c=freenode%23whatwg&s=10%20Feb%202016&e=10%20Feb%202016#c984850 for review.

PR #36, opened 2016-03-15
Fix #18: remove UTF-16 support from TextEncoder

### LINKED ISSUE #18: Consider removing TextEncoder support for UTF-16
AFAICT, non-UTF-8 encoders are exposed in three places in the Web Platform:

1. TextEncoder
2. Query string parsing for URL and the associated APIs
3. Form submission

URL parsing and form submission use UTF-8 when the encoding of the document is either of the UTF-16 encodings, AFAICT, TextEncoder is the only place where the Web Platform exposes UTF-16 encoder functionality.

This is pretty sad, if TextEncoder support UTF-16 just because it can and the support doesn't have strong use cases.

Am I wrong and do proper use cases exist for UTF-16 in TextEncoder to a degree that sets UTF-16 apart from the other legacy encoding that aren't supported in TextEncoder?

annevk:
This also allows us to remove the UTF-16 encoders as they are not
exposed in any way after TextEncoder support is removed.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-encoding.svg" width="100"></a></p>
 <h1>Encoding</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-12-february-2016">Living Standard — Last Updated 12 February 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-18-march-2016">Living Standard — Last Updated 18 March 2016</h2>
 
 <dl>
  <dt>Participate:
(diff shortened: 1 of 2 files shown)


inexorabletash:
Diff looks good.

We don't have stats on whether the TextEncoder constructor is typically called with or without the optional argument. But knowing that developers do like to fill in the blanks, it would surprise me if there weren't uses of `new TextEncoder("utf-8")` out there. Which means we'll never be able to use the first argument for anything else. I'm tempted to leave the optional argument it in there, with the "throw if not a utf-8 label" behavior.

domenic:
I really prefer not to have arguments that are not used for anything but for checking their value. That is a pattern that appears nowhere else on the platform. Instead, a simple comment in the source saying that we can't add a first argument to the constructor seems better.

annevk:
Ah yeah, I should add such a comment.

Ms2ger:
So `new TextEncoder("utf-16")` will silently change to use utf-8? Not sure how happy I am with that.

annevk:
Usage for that particular value is 0% so I'm not sure it matters.

egoroof:
> So new TextEncoder("utf-16") will silently change to use utf-8? Not sure how happy I am with that.

Yeah. It will cause silent bugs in existing code which uses utf16. Why not throw exception?

PR #38, opened 2016-03-31
Avoid teaching bad UTF-8 math.

hsivonen:
1. Instead of subtracting a magic number, do a bitwise AND to pick a number of least-significant bits that makes sense if you consider the UTF-8 bytes bitwise.
2. Instead of shifting by variable amounts in order to put all bits into their final destination directly, always shift by a constant six and let the higher bits be shifted a number of times.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-encoding.svg" width="100"></a></p>
 <h1>Encoding</h1>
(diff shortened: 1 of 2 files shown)


annevk:
Thank you. Would you be okay with you or me turning the "i.e." bits into paragraph notes? The existing encoding algorithms don't have any inline explanations.

hsivonen:
Works for me. I don't have the time to do it myself today, though, so please go ahead if you have the time.

annevk:
If you could review my fixup that'd be great. No rush though. (Note that it includes the generated file.) After that I'll squash, merge, and land on master.

hsivonen:
r+. Looks good. Thanks.

PR #39, opened 2016-04-01
Add additional legacy encodings

inexorabletash:
An improvement to our indexing and metrics analysis has revealed that we were incorrectly reporting on the use of some legacy encodings. A substantial number of web-exposed resources -- in particular, many generated with older hardware and exposed to the web through non-transcoding proxies -- use encodings not covered by Encoding standard. We believe it will be necessary to add support to browsers (and thus to the Encoding standard) to ensure this content remains accessible and usable.

The additional encodings are:

* cp037
* cp500
* cp875
* cp1026
* cp1140

(I was not able to get Anolis running; bit-rot appears to have set in. Someone will need to rebuild Overview.html using the Makefile after merging - sorry about that.)


annevk:
No worries about Anolis. I'll handle that.

1. encodings.json is not updated.
2. Labels need to be lowercase.
3. I wonder if there's a better name for single-byte encodings if we need this new variant.

And obviously this needs review from other implementers, such as @hsivonen. This is a pretty big change.

annevk:
Also, these have some serious security implications, since they are not ASCII-compatible. I'm sure how wise it would be to add these. We could reduce the implications a bit by not supporting the encoder, but that doesn't address all issues.

inexorabletash:
> 1. encodings.json is not updated.
> 2. Labels need to be lowercase.

Addressed in latest commit.

vyv03354:
Gecko has never supported EBCDIC encodings. I don't understand why those encodings are needed for Web compat unless Web==IE.

inexorabletash:
Now that April 1st is drawing to an end we can close this out. But first I'll quote the Jargon File:

> Hackers blanch at the very name of EBCDIC and consider it a manifestation of purest evil.

PR #41, opened 2016-04-06
Make the name of ISO-8859-8-I all upper case in encodings.json.

hsivonen:
Looks like a find-and-replace oversight.


annevk:
Thank you!

PR #43, opened 2016-04-12
Export the algorithms intended for use by other specs.

jyasskin:
Using the `<dfn>` format defined at
https://github.com/tabatkins/bikeshed/blob/master/docs/dfn-contract.md.

http://anolis.hoppipolla.co.uk/aquarium.py fails on this file because it contains a non-ascii comment (&#x01f4a9;), so I haven't tested that it actually produces the right output.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -855,7 +855,7 @@ to represent ranges.)
 <span title=concept-encoding-get>get an encoding</span> algorithm can be used first to turn a
 <span>label</span> into an <span>encoding</span>.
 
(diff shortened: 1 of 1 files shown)


domenic:
*Someone* should Bikeshed this spec.... also Fetch...

jyasskin:
Bikeshedding wouldn't have avoided this particular problem, since it also defaults `<dfn>`s to noexport. :)

annevk:
I added a couple more. Landed as https://github.com/whatwg/encoding/commit/bf42f4b898fe31b3c2d6cbd9684a907ea5cfee19. Thank you!

PR #45, opened 2016-04-19
Fix regression in utf-8 decoder

### LINKED ISSUE #42: utf-8 decoder lead byte switch branches 0xE0 to 0xEF, 0xF0 to 0xF4 need 'return continue'
https://github.com/whatwg/encoding/pull/38 changed how the lead byte is processed. It removed "Then (byte is in the range 0xC2 to 0xF4, inclusive) set UTF-8 code point to UTF-8 code point << (6 × UTF-8 bytes needed) and return continue."

The first two and final switch branches (0x00 to 0x7F, 0xC2 to 0xDF, and Otherwse) have returns, but the remaining two (0xE0 to 0xEF, 0xF0 to 0xF4) do not, so fall through to the trail byte processing.

Looks like they just need additional "Return continue" substeps added

annevk:
Caused by 19a25b5fcae895853d964b7ee6afa2fe9b6070a8. Fixes #42.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-encoding.svg" width="100"></a></p>
 <h1>Encoding</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-13-april-2016">Living Standard — Last Updated 13 April 2016</h2>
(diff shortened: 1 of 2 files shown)


annevk:
@inexorabletash can you review?

inexorabletash:
Looks correct.

It seems a bit odd to have the "return" in an unnumbered paragraph below the switch, but we don't have a formal grammar for specs yet.

annevk:
Yeah, really need to get on that some day. Have been thinking about it for a while now.

PR #55, opened 2016-06-01
Editorial: avoid upsetting lazy compilers

### LINKED ISSUE #49: Group constants together in half-width math
In the EUC-JP and Shift_JIS decoders, there's this math expression: "0xFF61 + byte − 0xA1".

Please rearrange it as "0xFF61 − 0xA1 + byte" to make it easy to get correct results without the implementor rearranging the expression.

In a programming language that's aware of integer overflow to the extent that the compiler doesn't perform constant folding in a way that would change whether an expression has integer overflow, the former doesn't allow constant folding but the latter does.

Furthermore, if the math happens to be done on the 16-bit precision, for example as an artifact of trying to compute a UTF-16 code unit, the expression in the spec actually has an overflow with the relevant values of _byte_. (Which doesn't matter e.g. in Java but does matter in e.g. debug-mode Rust.)

annevk:
Fixes #49.


hsivonen:
r+. Thanks.

PR #65, opened 2016-07-24
Editorial: make gb18030 math more clear

### LINKED ISSUE #63: Editorial: Clean up gb18030 math
Please clean up the math for 4-byte sequences in the gb18030 decoder to [perform the three multiplications by constant expressions and then adding the results](https://github.com/hsivonen/encoding-rs/blob/5668dc4b18bc6bab5a90b023b643e112c2fe5674/src/gb18030.rs#L143). This both helps readability and reduces [data dependencies](https://en.wikipedia.org/wiki/Data_dependency#Implications) if the spec text is transcribed to code.

Please clean up the math for 4-byte sequence in the gb18030 encoder to [use the `%` operator with constant expressions on the high-hand side](https://github.com/hsivonen/encoding-rs/blob/5668dc4b18bc6bab5a90b023b643e112c2fe5674/src/gb18030.rs#L302). This helps readability, reduces data dependencies if the spec text is transcribed to code and reduces operations when `/` and `%` with the same operands can be compiled into a single instruction.

annevk:
Fixes #63.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -9,7 +9,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-encoding.svg" width="100"></a></p>
 <h1>Encoding</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-29-july-2016">Living Standard — Last Updated 29 July 2016</h2>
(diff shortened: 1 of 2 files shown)


annevk:
Assigning to @hsivonen for review. Please don't merge if the date is no longer July 24, 2016. I'd prefer to fix the commit before landing in that case.

PR #79, opened 2016-11-10
Bikeshed conversion

annevk:


annevk:
@tabatkins I get these warnings that I'm not sure how to resolve:
```
LINK ERROR: Multiple possible 'dfn' local refs for 'encoding'.
Arbitrarily chose the one with type 'dfn' and for ''.
LINK ERROR: Multiple possible 'dfn' local refs for 'encode'.
Arbitrarily chose the one with type 'dfn' and for ''.
```
I think went through each usage of these definitions and annotated them with `for=/`, but to no avail.

inexorabletash:
If I replace the two uses of `>encode<` with `>encodex<` the errors disappear. Is Bikeshed treating 'encoding' and 'encode' as the same term via conjugation?

Maybe just add `lt=encode-hook` to those.

annevk:
I was thinking that might be it, but we cannot change the terms as other standards use them already. And that would be a misuse of lt.

inexorabletash:
Verified that's the cause (bikeshed/ReferenceManager.py, linkTextVariations). Not sure what to markup sprinkle to use, though.

annevk:
Ta, guess we'll wait for @tabatkins to weigh in then.

tabatkins:
Le sigh. This is an issue @franremy ran into before.  Y'all aren't doing anything wrong, let's see what I can do about this.

tabatkins:
Should be done now.

annevk:
Confirmed that I no longer get the errors. Thanks!

Issue #76, opened 2016-10-05
Editorial style nits


domenic:
Class names like Uint8Array, ArrayBuffer, TypeError, and RangeError should link to their definitions.

For associated "internal slots" (e.g. encoding, error mode, ignore BOM flag, etc.), their defining instance should be a `<dfn>`, and all subsequent uses of or references to them should be `<a>`s. Currently both are `<b>`.

annevk:
I don't see why those class names are different from DOMString. They're all defined by IDL.

domenic:
Agreed, DOMString should get linked too.

annevk:
Is there a Bikeshed issue on this? That should basically get this fixed.

annevk:
There's actually parts I could fix, but not the cross-referencing from IDL blocks.

annevk:
And the internal slot parts also still need to be fixed.

annevk:
If USVString really needs to be linked that would require an issue against Bikeshed itself. Everything else should be taken care off now. Thanks!

PR #81, opened 2016-11-11
Editorial: cross-reference more IDL terms

annevk:
Fixes part of #76.


annevk:
This needs https://github.com/tabatkins/bikeshed/issues/869 to land first since I'm not going to pretend "typedarray" is still a think by adding "webidl" to a defaults block.

Issue #47, opened 2016-05-02
index-jis0208.txt should be JIS X 0208 and add another index file


takahashim:
The CCS(coded character set) of index-jis0208.txt is CCS of "CP932" or "Windows-31J".  It is not JIS X 0208.
There are two differences:
- CP932 has more characters than JIS X 0208.
- All CCS in JIS X 0208 has its name, for mapping to UCS.  There are some difference between it and index-jis0208.txt.

The important role of JIS X 0208 is restriction of character set.  For example, some fonts in Japan have characters only in JIS X 0208, not CP932.  So we need a strict character set to implement converter like Shift_JIS encoder.  We need another index which is differ from index-jis0208.txt.

Another usecase: sometimes I want to convert Shift_JIS text into EPUB file. It's typical usecase of Shift_JIS decoder.  When I do it, I want to convert EM DASH in Shift_JIS (0x815C, 1-1-29) to EM DASH in Unicode (U+2014).  It's OK in JIS X 0208 (and JIS X 0213), but It's NG in CP932; the table of index-jis0208.txt maps EM DASH into HORIZONTAL BAR(U+2015).  It's not what I want to do.

So my suggestions is:
- index-jis0208.txt should be renamed, such as index-cp932 or index-windows31j,
- and should add another index same as JIS X 0208.

I have machine-readable mapping table, "JIS X 0213:2004 8-bit code vs Unicode mapping table"
http://x0213.org/codetable/jisx0213-2004-8bit-std.txt. But It's JIS X 0213, not JIS X 0208.
I found the issue #31 and I also don't have a real usecase of SHIFT_JISX0213 nor Shift_JIS-2004 (when I want to use characters in JIS X 0213, I always use UTF-8).  I need just only JIS X 0208, subset of jisx0213-2004-8bit-std.txt.

annevk:
The goal of this standard is to match browsers and similar clients, since they're the most dominant internet clients out there and whatever they do spreads elsewhere. Given that, I'm not sure such a change makes sense.

Also, if you're worried about fonts it would make sense to check the code points of the font against the code points you are looking to encode. I'm not sure how it makes sense to rely on the encoder.

vyv03354:
It would make sense to rename the current file to index-windows-932.txt because it will clarify what browsers and similar clients actually use. ("cp932" is ambiguous because IBM and Microsoft have different "cp932" definitions.) But I disagree any non-utf8 additions.

hsivonen:
[...]

PR #83, opened 2016-11-16
Note >8835 pointers in index jis0208 cannot be reached

annevk:
In EUC-JP and ISO-2022-JP encoders, when getting a pointer for a code
point in index jis0208, it’s important to note that the pointer is
always less than 8836. This is because the code points are duplicated
throughout the table and the index pointer returns the first.

Per a suggestion in #47.


inexorabletash:
lgtm

PR #84, opened 2016-11-16
Document an implementation buffer trick

annevk:
***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/annevk/implementation-considerations/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/ee2bd34...279eac3.html)

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -2700,6 +2700,10 @@ issues. Browsers also should disable this feature if the resource was decoded us
   byte in <a>gb18030</a> is identical.
 </ol>
 
+<p>The <a>ISO-2022-JP encoder</a> needs <a>ISO-2022-JP encoder state</a> as additional state, but
+other than that, none of the <a for=/>encoders</a> for <a for=/>encodings</a> in this standard
+require additional state or buffers.
+
 
 
 <h2 class=no-num id=acknowledgments>Acknowledgments</h2>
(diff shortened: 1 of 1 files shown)


annevk:
@hsivonen worth doing?

annevk:
Preview: https://encoding.spec.whatwg.org/branch-snapshots/annevk/implementation-considerations/#implementation-considerations

hsivonen:
I think this is worth doing, but I suggest highlighting that the single-byte buffer will always hold an ASCII value for gb18030 decoder and will always hold either 0x24 or 0x28 for the ISO-2022-JP decoder.

annevk:
@hsivonen I was going to add that encoders do not need an additional buffer at all, but what about the ISO-2022-JP encoder?

annevk:
(I added what the single-byte buffer would contain by the way.)

hsivonen:
> I was going to add that encoders do not need an additional buffer at all, but what about the ISO-2022-JP encoder?

The encoders are all stateless, except for the ISO-20222-JP encoder, and the only piece of state it needs is ASCII/Roman/JIS0208. No buffer needed.

(Additionally, the only reason the encoders need to process EOF is to transition ISO-2022-JP back into the ASCII state. Even the ISO-2022-JP encoder even without EOF signaling would produce a stream that decodes without errors.)

> (I added what the single-byte buffer would contain by the way.)

Thank you.

annevk:
I added a short paragraph on encoders.

hsivonen:
LGTM. Thank you.

PR #89, opened 2017-01-16
Non-normatively visualize the indexes

### LINKED ISSUE #78: Inform readers about the structure of the sparsity of index-euc-kr
index-euc-kr is rather sparse. That is, it has a lot of pointer space without corresponding characters.

To avoid making implementors ship bloated software and to avoid making implementors to work out the pattern of the index holes on their own, it would be courteous to add an informative note about the pattern of the holes in the index.

These holes are easier to see in [bestfit949.txt](http://www.unicode.org/Public/MAPPINGS/VENDORS/MICSFT/WindowsBestFit/bestfit949.txt).

Visually (I haven't verified programmatically), when lead is 0xC7 or greater, only trail bytes 0xA1...0xFE (inclusive) map to characters using only 94 slots of the 190 in each stride. When lead is less than 0xC7, trail bytes 0x5B...0x60 (inclusive) and 0x7B...0x80 (inclusive) are not used using only up to 178 slots of 190 in each stride.

Some strides (leads in the range 0xA2...0xAF) have even fewer than 178 slots used, but I haven't yet worked out if there's a pattern to those.

hsivonen:
Closes #78.

r? @annevk

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -68,6 +68,8 @@ if [ $BRANCH != "master" ] ; then
          > $BRANCH_DIR/index.html;
     cp *.txt $BRANCH_DIR/;
     cp *.json $BRANCH_DIR/;
+    cp *.css $BRANCH_DIR/;
+    python visualize.py $BRANCH_DIR/
     echo "Branch snapshot output to $WEB_ROOT/$BRANCHES_DIR/$BRANCH"
 else
     # Living standard, if master
@@ -76,6 +78,8 @@ else
          > $WEB_ROOT/index.html
     cp *.txt $WEB_ROOT/;
     cp *.json $WEB_ROOT/;
+    cp *.css $WEB_ROOT/;
+    python visualize.py $WEB_ROOT/
     echo "Living standard output to $WEB_ROOT"
 fi
 
--- a/deploy.sh
+++ b/deploy.sh
@@ -54,6 +54,8 @@ curl https://api.csswg.org/bikeshed/ -f -F file=@$INPUT_FILE -F md-status=LS-COM
      > $COMMIT_DIR/index.html;
 cp *.txt $COMMIT_DIR/;
 cp *.json $COMMIT_DIR/;
+cp *.css $COMMIT_DIR/;
+python visualize.py $COMMIT_DIR/;
 echo "Commit snapshot output to $WEB_ROOT/$COMMITS_DIR/$SHA"
 echo ""
 
@@ -69,7 +71,7 @@ if [ $BRANCH != "master" ] ; then
     cp *.txt $BRANCH_DIR/;
     cp *.json $BRANCH_DIR/;
     cp *.css $BRANCH_DIR/;
-    python visualize.py $BRANCH_DIR/
+    python visualize.py $BRANCH_DIR/;
     echo "Branch snapshot output to $WEB_ROOT/$BRANCHES_DIR/$BRANCH"
 else
     # Living standard, if master
@@ -79,7 +81,7 @@ else
     cp *.txt $WEB_ROOT/;
     cp *.json $WEB_ROOT/;
     cp *.css $WEB_ROOT/;
-    python visualize.py $WEB_ROOT/
+    python visualize.py $WEB_ROOT/;
     echo "Living standard output to $WEB_ROOT"
 fi
 
--- a/encoding.bs
+++ b/encoding.bs
@@ -656,20 +656,26 @@ changed, so has the <a>index</a>.
 
 <div class=note id=visualization>
  <p>There is a non-normative visualization for each index other than <a>index gb18030 ranges</a>.
- <a>index jis0208</a> also has an alternative <a>Shift_JIS</a> visualization.
- Additionally, there is visualization of the Basic Multilingual Plane coverage of each index other than <a>index gb18030 ranges</a>.
+ <a>index jis0208</a> also has an alternative <a>Shift_JIS</a> visualization. Additionally, there is
+ visualization of the Basic Multilingual Plane coverage of each index other than
+ <a>index gb18030 ranges</a>.
 
  <p>The legend for the visualizations is:
+
  <ul class=visualizationlegend>
   <li class=unmapped>Unmapped
   <li class=mid>Two bytes in UTF-8
-  <li class="mid contiguous">Two bytes in UTF-8, code point follows immediately the code point of previous pointer
+  <li class="mid contiguous">Two bytes in UTF-8, code point follows immediately the code point of
+  previous pointer
   <li class=upper>Three bytes in UTF-8 (non-PUA)
-  <li class="upper contiguous">Three bytes in UTF-8 (non-PUA), code point follows immediately the code point of previous pointer
+  <li class="upper contiguous">Three bytes in UTF-8 (non-PUA), code point follows immediately the
+  code point of previous pointer
   <li class=pua>Private Use
-  <li class="pua contiguous">Private Use, code point follows immediately the code point of previous pointer
+  <li class="pua contiguous">Private Use, code point follows immediately the code point of previous
+  pointer
   <li class=astral>Four bytes in UTF-8
-  <li class="astral contiguous">Four bytes in UTF-8, code point follows immediately the code point of previous pointer
+  <li class="astral contiguous">Four bytes in UTF-8, code point follows immediately the code point
+  of previous pointer
   <li class=duplicate>Duplicate code point already mapped at an earlier index
   <li class=compatibility>CJK Compatibility Ideograph
   <li class=ext>CJK Unified Ideographs Extension A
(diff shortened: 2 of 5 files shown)


hsivonen:
Oops. I forgot to actually call my `aria()` function.

annevk:
The reason Travis fails is because the visualization resources are not passing the HTML checker. Some are not NFC and some use code points U+0080 to U+009F. We'll either need to exclude them from HTML checking or address that somehow.

deploy.sh
```diff
@@ -68,6 +68,8 @@ if [ $BRANCH != "master" ] ; then
          > $BRANCH_DIR/index.html;
     cp *.txt $BRANCH_DIR/;
     cp *.json $BRANCH_DIR/;
+    cp *.css $BRANCH_DIR/;
+    python visualize.py $BRANCH_DIR/
```
annevk: Was it intentional to skip commit snapshots?

encoding.bs
```diff
@@ -18,6 +18,7 @@ Markup Shorthands: css off
 Translate IDs: dictdef-textdecoderoptions textdecoderoptions,dictdef-textdecodeoptions textdecodeoptions
 </pre>
 
+<style>@import 'visualization-colors.css';</style>
```
annevk: According to Tab we can use `<link>` and Bikeshed will move it correctly in the output.

encoding.bs
```diff
@@ -653,33 +654,66 @@ changed, so has the <a>index</a>.
 <var>code point</var> in <var>index</var>, or null if
 <var>code point</var> is not in <var>index</var>.
 
+<div class=note id=visualization>
+<p>There is a non-normative visualization for each index other than index gb18030 ranges.
```
annevk: Since this is inside a `<div>` it needs to be indented by one.

annevk: We should also xref the indexes here I think.

encoding.bs
```diff
@@ -653,33 +654,66 @@ changed, so has the <a>index</a>.
...
+index jis0208 also has an alternative Shift_JIS visualization.
+Additionally, there is visualization of the Basic Multilingual Plane coverage of each index other than index gb18030 ranges.
+
+<p>The legend for the visualizations is:
+<ul class=visualizationlegend>
+    <li class=unmapped>Unmapped
```
annevk: These have too much indentation. `<div>` + `<ul>` means two spaces.

encoding.bs
```diff
@@ -653,33 +654,66 @@ changed, so has the <a>index</a>.
...
 <p>These are the <a lt=index>indexes</a> defined by this
 specification, excluding <a>index single-byte</a>, which have their own table:
 
 <table>
- <tbody><tr><th colspan=2><a>Index</a><th>Notes
+ <tbody><tr><th colspan=4><a>Index</a><th>Notes
  <tr>
   <td><dfn export>index Big5</dfn>
   <td><a href=index-big5.txt>index-big5.txt</a>
+  <td><a href=big5.html>Visualization</a>
```
annevk: We should add a title here saying "index Big5 visualization". Even better would be to make each link text unique somehow since that affects screen readers.

encoding.bs
```diff
@@ -653,33 +654,66 @@ changed, so has the <a>index</a>.
...
   <td>This matches the Big5 standard in combination with the
   Hong Kong Supplementary Character Set and other common extensions.
  <tr>
   <td><dfn export>index EUC-KR</dfn>
   <td><a href=index-euc-kr.txt>index-euc-kr.txt</a>
+  <td><a href=euc-kr.html>Visualization</a>
+  <td><a href=euc-kr-bmp.html>BMP coverage</a>
   <td>This matches the KS X 1001 standard and the Unified Hangul Code, more
-  commonly known together as Windows Codepage 949.
+  commonly known together as Windows Codepage 949. This index covers the Hangul Syllables
```
annevk: s/This index/It/

deploy.sh
hsivonen: No. I just didn't understand the purpose of that directory correctly.

encoding.bs
hsivonen: OK.

encoding.bs
hsivonen: OK

encoding.bs
hsivonen: OK

encoding.bs
hsivonen: I did the latter.

encoding.bs
hsivonen: OK.

hsivonen:
> Some are not NFC 

Prepended with space.

> and some use code points U+0080 to U+009F.

Replaced with SVG.

visualization.css
```diff
@@ -0,0 +1,92 @@
+@import 'https://resources.whatwg.org/fonts/faces.css';
+@import 'visualization-colors.css';
+
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+
+/* Common styling from standard.css */
+h1 {
+  color: ##3c790a; /* WHATWG Green */
```
zcorpan: Double #

visualize.py
```diff
@@ -0,0 +1,263 @@
...
+        label += ", duplicate"
+    return label
+
+
+def format_code_point(code_point):
+    if code_point >= 0x80 and code_point < 0xA0:
+        # HTML prohibits C1 controls
+        # TODO draw some fancy SVG hex inside the square
+        return "<svg><rect x=1 y=1 width=14 height=14 stroke=black stroke-width=2 fill=none /></svg>"
+    as_str = unichr(code_point)
```
zcorpan: I get:
```
Traceback (most recent call last):
  File "visualize.py", line 258, in <module>
    format_index(name, row_length, lang, bmp, duplicates, byte_rule)
  File "visualize.py", line 206, in format_index
    out_file.write((u"<td class='%s %s%s%s' aria-label='%s'><dl><dt>%d<dd lang=%s>%s<dd>U+%04X</dl>" % ("contiguous" if contiguous else "discontiguous", classify(code_point), " duplicate" if duplicate else "", check_compatibility(code_point), aria(code_point, contiguous, duplicate), pointer, lang, format_code_point(code_point), code_point)).encode("utf-8"))
  File "visualize.py", line 124, in format_code_point
    as_str = unichr(code_point)
ValueError: unichr() arg not in range(0x10000) (narrow Python build)
```

...with Python 2.7.10

hsivonen: Python's sad Unicode support strikes again. How did you get your copy of Python? Apple?

The script works on Python 2.7.12 shipped on Ubuntu 16.04. Debian and, therefore, Ubuntu ships wide Python. It seems that Travis also runs Ubuntu, so it should be OK there.

@annevk, Is making the script work on e.g. Apple-shipped Python a blocker for merging?

visualization.css
hsivonen: Thanks. Fixed in the latest push.

visualize.py
zcorpan: I *think* I got it from Apple, but I don't really know.

gsnedders: Apple and some OS X package managers ship narrow builds. If the issue is literally just `unichr`, then it is at least easily worked around.

hsivonen: > If the issue is literally just `unichr`, then it is at least easily worked around. 

Yeah, I guess tomorrow I'll be writing the wrapper (Unicode string for code point regardless of 16 vs. 32-bit units) that Python should have had the good sense to put in the standard library.

hsivonen: It now works with Apple-shipped Python, too. (Actually tested on a Mac and diffed the output with Ubuntu's wide Python output.)

zcorpan:
Builds for me now for `./deploy.sh --local`. 👍

annevk:
https://travis-ci.org/whatwg/encoding/builds/193654621 still reports PUA and NFC warnings. I think the NFC warnings we have to accept per conversation on IRC. Does the same go for PUA? It would be good to mention in the commit message what is expected there.

annevk:
One other problem I noticed is that the wrapping is not following 100 columns which is what we're starting to adopt everywhere (though no linebreaks inside "inline" elements). I'm happy to fix that once you consider this ready.

hsivonen:
> Does the same go for PUA?

As noted on IRC, yes. For example, on Ubuntu, showing PUA as PUA gives the insight that the PUA code points in the left part of the last row of gb18030 show up as radicals that fit reasonably between the non-PUA radical code points around them. Then one can go read Lunde and learn why this is so: the PUA radicals weren't mapped in Unicode at the time gb18030 was first specced.

> One other problem I noticed is that the wrapping is not following 100 columns which is what we're starting to adopt everywhere (though no linebreaks inside "inline" elements). I'm happy to fix that once you consider this ready.

I think this is ready. Thanks.

annevk:
Okay, I'll go fix that. Could you write a commit message that explains the change and the warnings? Just post it as a comment and I'll use it when squash and merging.

hsivonen:
Add visualizations for the indexes

PUA and NFC validation warnings are expected. Exposing the PUA code points as PUA in HTML is useful to seeing what those code points map to (if to anything) in system fonts, which may give insights about their usage. Exposing singletons (compatibility ideographs and scientific units) without normalizing them is useful for having browser "Find" functionality match whatever you get as output of a converter you might be developing.

annevk:
Thanks @hsivonen, they look great and will surely help folks understand how these encodings work!

domenic:
May I suggest a WHATWG blog post explaining this space and the new visualizations to people? :)

PR #98, opened 2017-03-19
Editorial: enable indexes and fix a couple minor nits

annevk:
***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/annevk/indexing/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/ee2bd34...ec3829c.html)


annevk:
Apart from indexes a lot of this is making casing more consistent (ISO-8859- rather than iso-8859-) and quite a bit is just in comments. Should carefully check again though before landing.

PR #100, opened 2017-03-19
Editorial: remove floor() operation in favor of integer division

### LINKED ISSUE #66: "In equations, all numbers are integers, additio..."
https://encoding.spec.whatwg.org/#terminology

> In equations, all numbers are integers, addition is represented by "+", subtraction by "−", multiplication by "×", division by "/", calculating the remainder of a division (also known as modulo) by "%", logical left shifts by "&lt;&lt;", logical right shifts by "&gt;&gt;", bitwise AND by "&amp;", and bitwise OR by "|". floor(x) is the largest integer not greater than x.

"all numbers are integers" contradicts "floor(x) is the largest integer not greater than x". To be more specific, it implies that x is an integer and therefore floor(x) is a no-op.

I suggest defining "/" as integer division and dropping all mentions of floor() from the spec. floor(pointer / 10 / 126 / 10) should be changed to pointer / (10 \* 126 \* 10) to avoid loss of precision, and similarly with floor(pointer / 10 / 126).

annevk:
Fixes #66.

***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/annevk/roof/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/ee2bd34...7eba85a.html)

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -101,10 +101,9 @@ many reasons that is now the mandatory encoding for all things.
 <p>Hexadecimal numbers are prefixed with "0x".
 
 <p>In equations, all numbers are integers, addition is represented by "+", subtraction by "&minus;",
-multiplication by "×", division by "/", calculating the remainder of a division (also known as
-modulo) by "%", logical left shifts by "&lt;&lt;", logical right shifts by ">>", bitwise AND by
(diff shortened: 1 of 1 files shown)


hsivonen:
LGTM. Thanks.

Issue #97, opened 2017-03-18
Important warnings getting lost in the noise


annevk:
I was afraid of this. Look at the output of https://travis-ci.org/whatwg/encoding/builds/204200335 and try to spot the problem. It's #95.

@sideshowbarker how hard is it to make it possible to configure warnings? Alternatively, maybe we can instrument separate runs for index.html and the other resources whereby index.html doesn't tolerate warnings.

sideshowbarker:
The command-line checker and the network API to the checker currently don’t provide any way to configure which warnings get emitted. The only option is to show all warnings along with errors, or else to show only the errors and show no warnings.

Anyway I think the particular noise problem at https://travis-ci.org/whatwg/encoding/builds/204200335 would be a non-problem if I just made the checker quit emitting the “Text run is not in Unicode Normalization Form C“ warning.

That message used to be an error but years ago I demoted it to being a warning. See https://bugzilla.validator.nu/show_bug.cgi?id=838. The reason I demoted it that the HTML spec does not any normative requirements at all that text being in Unicode Normalization Form C—no “must” requirements and not even any “should” requirement. In fact I think the HTML spec doesn’t mention Unicode Normalization Form C at all.

So the reason I have the checker emitting any message about Unicode Normalization Form C isn’t because of any requirements in the HTML spec but I think instead because of Internationalization best-practice guidance published elsewhere. But in that case it would also be something that the W3C Internationalization Checker https://validator.w3.org/i18n-checker/ is already checking.

And so as far as the HTML checker goes, that warning is a discretionary warning (not required by anything in the HTML spec) and we can use our own discretion to decide it should just be dropped from the HTML checker.

So do think we should just drop it or not? I guess maybe these generated files for the Encoding spec are kind of a special case but I now wonder if maybe they’re not and if other people trying to check documents in other encodings run into this same problem (that is, all errors and warnings about more-serious problems just get drowned out in all the noise from the “Text run is not in Unicode Normalization Form C“ warnings.

aphillips:
That warning is an i18n best practice, but no longer something i18n promotes as a requirement.

That said, it is likely to be specific to Encoding (or to i18n specs: I see it regularly), but it will be rare (and undesirable to ignore) in most normal ones. Might be nice to emit once? Or have an option to suppress it for folks writing specs with lots of Unicode examples?

annevk:
Yeah, I think typically you want to know about non-NFC code points.

r12a:
[...]

PR #101, opened 2017-04-10
Use HTML checker network API + suppress warnings

sideshowbarker:
* Switch to using the HTML checker network API rather than `vnu.jar`
* Add a `CHECKER_PARAMS` environment variable to `.travis.yml` to cause the
  HTML checker to suppress warnings about Unicode non-NFC text runs and also
  about use of Unicode Private Use area, and about missing `html[lang]`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,11 @@ language: generic
 
 env:
   global:
-    - ENCRYPTION_LABEL: "eaa0f98d3fae"
-addons:
-  apt:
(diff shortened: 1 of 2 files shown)


annevk:
This doesn't quite work. You need to modify the local `deploy.sh` script. This script hasn't been generalized yet (@domenic has been planning to add support for files that only need to generated sometimes).

annevk:
This fixes #97.

Issue #59, opened 2016-06-20
EUC-jp encoding/decoding support


r12a:
Results for a series of tests for EUC-jp encoding/decoding can be found at 
https://www.w3.org/International/tests/repo/results/encoding-dbl-byte.en#eucjp

The tests can be run from that page (select the link in the left-most column) or get the tests from the WPT repo.  There is a PR at
https://github.com/w3c/web-platform-tests/pull/3198

The tests check whether:
1. the browser produces the expected byte sequences for all characters in the euc-jp encoding after 0x9F when encoding bytes for a URL produced by a form, using the encoder steps in the specification.
2. the browser produces percent-escaped character references for a URL produced by a form when encoding miscellaneous characters that are not in the euc-jp encoding. (tests for several ranges)
3. same two types of test when writing characters to an href value
4. the browser decodes all characters as expected from a file generated by encoding all pointers in the euc-jp encoding per the encoder steps in the specification.
5. the browser decodes characters that are not recognised from the euc-jp encoding as replacement characters.

The following summarises the current situation according to my testing, for major desktop browsers.  (I will be adding nightly results and perhaps other browsers in time.)  The table lists the number of characters that were NOT successfully converted by the test.

![screen shot 2016-06-20 at 16 53 26](https://cloud.githubusercontent.com/assets/4839211/16201008/8e5ab82e-3707-11e6-9cad-35abc08d3204.png)

Notes:
- Edge fails all href encode tests because characters are not converted to percent-escapes in the href attribute.
- Firefox fails all href encode tests for characters not in the encoding because it converts characters to percent-escaped Unicode values instead.
- eucjp-decode-index: Edge fails on all and only the JIS-X-0212 characters, because it doesn't recognise 0xAF as the first in a 3-byte sequence.

Can we please investigate the failures to ascertain whether:
1. the browser needs to be changed
2. the spec needs to be changed
3. the test is at fault

The following tool may be helpful for investigating issues. It converts between byte sequences and characters for all encodings in the Encoding spec. http://r12a.github.io/apps/encodings/

r12a:
List of bugs raised:

- https://bugs.webkit.org/show_bug.cgi?id=159889
- https://bugzilla.mozilla.org/show_bug.cgi?id=1285399
- https://bugs.chromium.org/p/chromium/issues/detail?id=626400
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8202215/

jungshik:
[...]

PR #103, opened 2017-04-28
EUC-JP: only unwind ASCII bytes

annevk:
See https://github.com/whatwg/encoding/issues/59#issuecomment-297750352
for context.

***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/annevk/euc-jp-ascii/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/612f9c6...b3bbc1b.html)

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -1923,8 +1923,7 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
 
    <li><p>Unset the <a>EUC-JP jis0212 flag</a>.
 
-   <li><p>If <var>byte</var> is not in the range 0xA1 to 0xFE, inclusive,
-   <a>prepend</a> <var>byte</var> to
+   <li><p>If <var>byte</var> is an <a>ASCII byte</a>, <a>prepend</a> <var>byte</var> to
    <var>stream</var>.
 
    <li><p>If <var>code point</var> is null, return <a>error</a>.


annevk:
@sideshowbarker the build process ended up timing out doing the validation. I wonder if we should actually detect which files changed and only do validation for those.

encoding.bs
```diff
@@ -1923,8 +1923,7 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
 
    <li><p>Unset the <a>EUC-JP jis0212 flag</a>.
 
-   <li><p>If <var>byte</var> is not in the range 0xA1 to 0xFE, inclusive,
-   <a>prepend</a> <var>byte</var> to
+   <li><p>If <var>byte</var> is an <a>ASCII byte</a>, <a>prepend</a> <var>byte</var> to
```
hsivonen: This should say:
If code point is null and byte is an ASCII byte...

(As in the EUC-KR case.)

Or, better yet, reorder the steps so that code point gets returned early if it is non-null and then the "prepend" and "return error" steps don't need to be qualified with "if code point is null".

annevk: Oh, thanks, I didn't realize this had two bugs.

annevk: I don't see why we have to check that code point is null. We already know it's null because byte is an ASCII byte. (That doesn't work with EUC-KR because ASCII byte is inside the valid range.)

I would be okay with changing order though, I guess I should look at how much that would affect.

annevk: If we only change null to non-null checks when it would affect at least two null checks, we'd have to change gb18030, Big5, EUC-JP (though per above it doesn't really need a null check), Shift_JIS, and EUC-KR.

I'm okay either way.

hsivonen: I think the definitions would become more readable if there was an early return with code point if code point is not null and then as ASCIIness/prepend check and then return error.

But you are right that in the case of EUC-JP, the trail can't be in the ASCII range, so maybe it's best to land this without the null check reordering and then do the null check  reordering as a separate editorial-only changeset.

annevk: I will do that.

annevk: #107 is the follow-up.

PR #104, opened 2017-04-28
Editorial: ensure various bits link properly

annevk:
***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/annevk/editorial/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/9ce8d47...fcf3484.html)

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -624,7 +624,7 @@ form submission, which both need exactly this.
 
 <h2 id=indexes>Indexes</h2>
 
-<p>Most legacy <a for=/>encodings</a> make use of an <dfn>index</dfn>. An
+<p>Most legacy <a for=/>encodings</a> make use of an <dfn id=index>index</dfn>. An
 <a>index</a> is an ordered list of entries, each entry consisting of a pointer and a
 corresponding code point. Within an <a>index</a> pointers are unique and code points can be
 duplicated.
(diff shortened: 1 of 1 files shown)


annevk:
@tabatkins I have a dfn "index" with ID "index" and this clashes with the newly added "Index" section which also wants ID "index" and is auto-generated by Bikeshed. I tried using Translate IDs, but that just ends up replacing both, which is later followed by deduping, rather than using Translate IDs only for the IDs generated by Bikeshed. Is that a bug or am I doing something wrong here?

domenic:
My two cents: although it does sound like this scenario wasn't considered while designing Bikeshed, and thus is a little wacky, I think it's better to not define something with the ID "index", so that in every Bikeshed spec `#index` goes to the same thing (the index of terms). `concept-index` or similar might be the way to go.

annevk:
@tabatkins ping.

@domenic I guess I could rename it too, but that seems bad if someone uses that fragment today.

tabatkins:
Hmm, I can definitely see the confusion with how `Translate IDs` is defined ("maps Bikeshed-generated IDs to manually-specified IDs") and how it actually works (attempts to translate all and only definitions, headings, issues, and examples).

I should fix this to do what it actually says on the tin.

tabatkins:
And I just pushed. Using it on "index" should work now, and only rename the Index ID, not your dfn with a manually-specified ID.

annevk:
Thanks!

PR #109, opened 2017-05-07
Meta: retry HTML checker by curl, suppress warnings

sideshowbarker:
This restores 9ce8d47 but refines such that if curl hits a timeout, it
will automatically retry each request up to 7 times.

From the curl(1) man page:

> --retry <num>
>    If  a  transient  error is returned when curl tries to perform a
>    transfer, it will retry this number of times before  giving  up.
>    Setting  the  number to 0 makes curl do no retries (which is the
>    default). Transient error means either: a timeout,  an  FTP  4xx
>    response code or an HTTP 5xx response code.
>
>    When  curl  is about to retry a transfer, it will first wait one
>    second and then for all forthcoming retries it will  double  the
>    waiting  time until it reaches 10 minutes which then will be the
>    delay between the rest of the retries.

Addresses #97

***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/sideshowbarker/checker-via-curl-retry/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/eedd518...19aa804.html)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,10 @@ language: generic
 
 env:
   global:
-    - ENCRYPTION_LABEL: "eaa0f98d3fae"
-addons:
-  apt:
-    packages:
-      - oracle-java8-set-default
-install:
-  - curl -O https://sideshowbarker.net/nightlies/jar/vnu.jar
+    - ENCRYPTION_LABEL="eaa0f98d3fae"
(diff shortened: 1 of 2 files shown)


annevk:
Did you try manipulating visualize? I wonder if it ends up completing reliably now.

sideshowbarker:
> Did you try manipulating visualize? I wonder if it ends up completing reliably now.

OK yeah just tried in https://travis-ci.org/whatwg/encoding/builds/235209188 and it worked as expected

sideshowbarker:
OK I think this now correctly handles all the cases we discussed and is ready to be final review and to be merged if it all looks right otherwise. Sorry for all the noise while debugging it

PR #111, opened 2017-05-11
gb18030 decoder: unwind from fourth byte when it's not a digit

### LINKED ISSUE #110: Should not prepend gb18030 second and third when range decode fails upon fourth
Consider a [run of byte pairs](https://hsivonen.com/test/moz/gb18030-alignment.htm) where each pair starts with a byte that's valid as first or third and the second byte is valid as second or fourth in a four-byte gb18030 sequence.

The first two pairs form a four-byte sequence that fails the gb18030 ranges lookup. The spec, as written, realigns the decoder such that the second pair of the first out-of-range four byte sequence is next treated as the first half of another four-byte sequence.

The test case linked to above decodes to four replacement characters in Firefox, Chromium and into four question marks in IE and Edge. (Didn't test Safari.) However, per spec, the run decodes into "�9¶ĭ¶�9".

One error causing the decoder to misalign such that subsequent non-error characters come into existence where there previously were none in browsers can't be good.

Suggested fix:
When processing the fourth byte in a sequence, first check it it is in the range 0x30 to 0x39 inclusive. If it isn't, prepend the second, third and fourth bytes in the sequence per current spec and return error.

Then set code point to the index gb18030 ranges code point. If code point is null, prepend only the fourth byte and return error.

This would deviate from the current browser behavior by unmasking the last ASCII byte in the sequence the way we always unmask failed ASCII in two-byte sequences. However, unlike in the current spec, the ASCII byte that's second in the sequence would get swallowed into a REPLACEMENT CHARACTER. This is fine, because we know that ASCII byte had non-ASCII bytes before and after.

(Trying to emit two errors with an ASCII character in between as one step would be a novel behavior for the spec and would totally mess up the assumptions I've made about error emission in implementation.)

annevk:
Instead of always unwinding if there’s no code point when consuming the
fourth byte, only unwind when the fourth byte is not an ASCII digit.
This does mean that ASCII digits can be masked, but since ASCII digits
are not used as delimiter in any format this is highly unlikely to be
used in any attacks (and also matches existing implementations better).

Fixes #110.

***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/annevk/gb18030-fourth-byte/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/378393f...951c471.html)

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -1641,11 +1641,13 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
    <li><p>Set <a>gb18030 first</a>, <a>gb18030 second</a>, and
    <a>gb18030 third</a> to 0x00.
 
-   <li><p>If <var>code point</var> is null,
-   <a>prepend</a> <var>buffer</var> to
-   <var>stream</var> and return <a>error</a>.
+   <li><p>If <var>code point</var> is non-null, return a code point whose value is
+   <var>code point</var>.
 
-   <li><p>Return a code point whose value is <var>code point</var>.
+   <li><p>If <var>byte</var> is not in the range 0x30 to 0x39, inclusive, <a>prepend</a>
+   <var>buffer</var> to <var>stream</var>.
+
+   <li><p>Return <a>error</a>.
   </ol>
 
  <li>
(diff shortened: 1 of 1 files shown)


hsivonen:
Works for me, but, editorially, I'd still prefer 0x30 to 0x39 checked only once and an early return before the range decode if the byte isn't in 0x30 to 0x39 inclusive.

annevk:
Rewrote it.

hsivonen:
LGTM. Thank you.

PR #117, opened 2017-07-04
Add "replacement" as label for the replacement encoding

### LINKED ISSUE #70: Add "replacement" as a label for the replacement encoding
I've written "The name of an encoding is also one of its labels, except in the case of the replacement encoding whose name is not one of its labels." inconveniently many times when either [documenting code that implements the Encoding Standard](https://hsivonen.fi/rs/encoding_rs/struct.Encoding.html) or when [otherwise explaining the concepts](https://bugzilla.mozilla.org/show_bug.cgi?id=1191841#c15).

Also, I've written code that does something like "if input string is 'replacement', don't run _get an encoding_, otherwise, run _get an encoding_" when working with [interfaces](https://dxr.mozilla.org/mozilla-central/source/intl/uconv/nsScriptableUConv.cpp#270) [that](https://dxr.mozilla.org/mozilla-central/source/netwerk/base/nsUnicharStreamLoader.cpp#177) [were](https://dxr.mozilla.org/mozilla-central/source/netwerk/base/nsStandardURL.cpp#221) [designed](https://hg.mozilla.org/integration/mozilla-inbound/rev/e07157e29deedd1b2bda60c62f307503c558844f) (four links there, but GitHub's styling makes it unobvious) before there was clarity of what strings are _labels_ and what strings are _names_ used where an enum value or a reference to a singleton object representing an encoding would be more appropriate software design (when those interfaces potentially have callers in add-ons that I can't fix).

All this would become simpler if _get an encoding_ for the _name_ of an encoding always returned the encoding itself. However, it's kinda sad to expose another Web-exposed label just to make implementing and explaining stuff easier, so I'm not sure if I should request this.

But at least this deserves some discussion.

annevk:
Fixes #70.

***
[Preview](https://encoding.spec.whatwg.org/branch-snapshots/replacement/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/encoding/eedd518...25d58aa.html)

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -582,12 +582,13 @@ prescribes, as that is found to be necessary to be compatible with deployed cont
  <tbody>
   <tr><th colspan=2><a href=#legacy-miscellaneous-encodings>Legacy miscellaneous encodings</a>
   <tr>
-   <td rowspan=5><a>replacement</a>
+   <td rowspan=6><a>replacement</a>
    <td>"<code>csiso2022kr</code>"
   <tr><td>"<code>hz-gb-2312</code>"
   <tr><td>"<code>iso-2022-cn</code>"
   <tr><td>"<code>iso-2022-cn-ext</code>"
   <tr><td>"<code>iso-2022-kr</code>"
+  <tr><td>"<code>replacement</code>"
   <tr>
    <td><a>UTF-16BE</a>
    <td>"<code>utf-16be</code>"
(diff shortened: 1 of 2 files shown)


hsivonen:
It would be good to non-normatively remark in the "Names and labels" section: "For each encoding, ASCII-lowercasing its name yields one of its labels."

r+ with or without that addition.

PR #121, opened 2017-09-03
Meta: use the common deploy.sh

foolip:


foolip:
This successfully deployed a branch snapshot to the new server. I'm not sure whether the checker step is equivalent before and after since the method used is different. @sideshowbarker?

annevk:
Would the output from the checker even show if it's non-fatal with the current setup? It seems like it might not be shown, which isn't ideal.

annevk:
https://travis-ci.org/whatwg/encoding/builds/232362067 is from before @sideshowbarker added the filter. Shows lots of warnings "Text run is not in Unicode Normalization Form C." that are typically useful to know about, but we know are not a problem for this document, which is why we filter them.

annevk:
(I'd actually argue for making such warnings an error for WHATWG documents, except when we decide it's not an error.)

foolip:
https://travis-ci.org/whatwg/encoding/builds/271487608 is the run with these changes, and the checker must have run and passed, even though there's no output. So I guess the common deploy.sh is more lenient than the previous setup? But https://checker.html5.org/?doc=https%3A%2F%2Fencoding.spec.whatwg.org%2F shows no errors, is that expected?

annevk:
Oh, then the problem is that you don't run the checker on all the other HTML files that end up being generated.

foolip:
@sideshowbarker, the common deploy.sh does `/usr/lib/jvm/java-8-oracle/jre/bin/java -jar vnu.jar --skip-non-html "$WEB_ROOT"`, doesn't that traverse and check all HTML files?

sideshowbarker:
> the common deploy.sh does `/usr/lib/jvm/java-8-oracle/jre/bin/java -jar vnu.jar --skip-non-html "$WEB_ROOT"`, doesn't that traverse and check all HTML files?

Yes, that traverses the entire `$WEB_ROOT` tree and runs the checker on every HTML file it finds

sideshowbarker:
> I'm not sure whether the checker step is equivalent before and after since the method used is different.

I’m not sure either. Is the difference that in case it’s sending the documents to https://checker.html5.org/ for checking using checker’s network API, and in the other case it’s downloading the jar and checking the documents locally?

If so, I think we want to be using the checker’s network API—not the jar—for all specs except the HTML spec. See https://github.com/whatwg/meta/issues/4#issuecomment-275128648 for why.

annevk:
Well, Encoding also has a lot of HTML files and I suspect they may end up being rather big in total, but I haven't verified.

The main problem here is that @foolip removed filtering for warnings, but the new output doesn't show warnings. So where did those go?

foolip:
OK, so I've run the existing script locally a bit to understand what these warnings are. https://encoding.spec.whatwg.org/big5.html has warnings like "Text run is not in Unicode Normalization Form C.", which show up when using the https://checker.html5.org/ API. The warnings also show up when using `java -jar vnu.jar encoding.spec.whatwg.org/big5.html` locally, *but* the exit code is 0.

@sideshowbarker, is there a way of making errors into warnings and the suppressing some warnings using the vnu.jar?

foolip:
Hmm, looks like `--filterfile` and `--filterpattern` are the way to suppress, but how to make warnings errors?

sideshowbarker:
There previously was no way of making the checker exit non-zero if there are only warnings, but I just today added a new `--Werror` option, so it’s in the current jar:

https://sideshowbarker.net/nightlies/jar/vnu.jar

foolip:
Yay, thanks! I think we should add a CHECKER_FILTER environment variable that is passed to `--filterpattern`, and then this will be unblocked. Although, I don't think I'll try to retain the logic that's clever about which files to validate, so deploys will start taking longer. @annevk, is that acceptable?

foolip:
This is ready for review again, best reviewed together with https://github.com/whatwg/whatwg.org/pull/111.

Issue #69, opened 2016-08-19
TextEncoder#encode - write to existing Uint8Array


maciejhirsz:
For binary protocols, such as msgpack, performance of writing/reading strings can be crucial, however those strings rarely (virtually never) exist on a binary buffer on their own. When decoding a string from a binary buffer of Uint8Array (obtained over, say, websockets) it's possible to cheaply create a subarray and pass it over to `TextDecoder#decode`, as such the decoding story is pretty straight forward.

The writing story is much more complicated.

In order to write a string into an existing buffer, one must take the result `Uint8Array` from `TextEncoder#encode` and `TypedArray#set` it at a desired offset into another buffer, which introduces unnecessary copying and, as result, is slower than plain JavaScript encoding implementations (at least for UTF-8).

A much better solution would be analog to the Node.js [`Buffer#write` implementation](https://nodejs.org/api/buffer.html#buffer_buf_write_string_offset_length_encoding) where the encoder can write a string to a buffer at a desired offset. Example signature could be something like:

```
TextEncoder#write(DOMString source, Uint8Array target [, Number offset [, Number length]]) -> Number bytesWritten
```

annevk:
Thank you for reporting this, I've heard a similar request aired for `fetch()` at some point and there might be other APIs where folks want buffer reuse. I wonder if we should adopt some kind of pattern for this.

@domenic? @inexorabletash?

inexorabletash:
Yep, legit request we've heard from potential users before. I don't know what the pattern should be like, or if it should be done at this level (new/extended method) or if we should add Stream support and there should be a generic stream-into-buffer mechanism.

With the `write()` proposal above, we'd want to return the amount of data consumed (since the encoder won't buffer arbitrary amounts itself) and amount written. That also sounds like a common element of this sort of pattern. (And also the kind of thing streams are good at, so hoping there's a generic solution.)

domenic:
Right, if we get transform streams for text encoding, then you'd be able to use a BYOB reader to target specific segments of your destination buffer.

A lower-level method might make sense. If so, don't have separate offset/length parameters; those are already part of the Uint8Array (which is a { byteOffset, byteLength, buffer } tuple).

shaunc:
To my mind writing to/reading from an ArrayBuffer would be a better interface. Alternately `DataView.getString/setString` methods could be added in conjunction with this interface.

hsivonen:
[...]

PR #127, opened 2017-12-11
Make TextEncoder and TextDecoder be transform streams

### LINKED ISSUE #72: Add Streams support
In https://github.com/whatwg/encoding/commit/9224c4cad81fd5720306fe79241cf04b5526a3f0#commitcomment-19169480 @jakearchibald @tyoshino @ricea are having a discussion about adding https://streams.spec.whatwg.org/ to the Encoding Standard that I'd prefer us to continue in this issue (and the eventual pull request). That will make it easier to track for other folks participating in this repository. Hope that's okay.

ricea:
Add `readable` and `writable` attributes to TextEncoder and TextDecoder
objects to permit them to be used to transform ReadableStreams via the
`pipeThrough()` method.

This integrates the Encoding Standard with the Streams Standard. See
https://streams.spec.whatwg.org/#ts-model for the definition of a
"transform stream" and https://streams.spec.whatwg.org/#rs-pipe-through
for an explanation of the `pipeThrough()` method.

A TextEncoder object can be used to transform a stream of strings to a
stream of bytes in UTF-8 encoding. A TextDecoder object can be used to
transform a stream of bytes in the encoding passed to the constructor to
strings.

The implementation delegates to a TransformStream object internally,
which provides the glue logic that ties the `readable` and `writable`
together.

There is a human-readable version of these changes at
http://htmlpreview.github.io/?https://github.com/ricea/encoding-streams/blob/master/patch.html

There is a prollyfill and tests for the new functionality at
https://github.com/GoogleChromeLabs/text-encode-transform-prollyfill

Closes #72.

***
<a href="https://whatpr.org/encoding/127.html" title="Last updated on Mar 30, 2018, 10:29 AM GMT (c39d7a1)">Preview</a> | <a href="https://whatpr.org/encoding/127/f381389...c39d7a1.html" title="Last updated on Mar 30, 2018, 10:29 AM GMT (c39d7a1)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -1540,8 +1540,8 @@ must run these steps:
 
  <p class=note>{{DOMString}} is used here so that a surrogate pair that is split between chunks can
  be correctly reassembled into the appropriate code point in the output. The behaviour is otherwise
- identical to {{USVString}}. In particular, replacement characters will be used where necessary to
- make the output valid UTF-8.
+ identical to {{USVString}}. In particular, lone surrogates will be replaced with U+FFFD so that the
+ output is always well-formed UTF-8.
 
  <li><p>Convert <var>input</var> to a <a for=/>stream</a>.
 
--- a/encoding.bs
+++ b/encoding.bs
@@ -1314,6 +1314,8 @@ method, when invoked, must run these steps:
   </ol>
 </ol>
 
+<hr>
+
 <p>The <dfn id=concept-td-decode-and-enqueue>decode and enqueue a chunk</dfn> algorithm, given a
 {{TextDecoder}} <var>dec</var> and a <var>chunk</var>, runs these steps:
 
@@ -1528,6 +1530,8 @@ must run these steps:
   </ol>
 </ol>
 
+<hr>
+
 <p>The <dfn id=concept-te-encode-and-enqueue>encode and enqueue a chunk</dfn> algorithm, given a
 {{TextEncoder}} <var>enc</var> and <var>chunk</var>, runs these steps:
 
@@ -1570,7 +1574,6 @@ must run these steps:
  </ol>
 </ol>
 
-
 <p>The <dfn id=concept-te-convert-code-unit-to-scalar-value>convert code unit to scalar value</dfn>
 algorithm, given a {{TextEncoder}} <var>enc</var>, <var>token</var> and <var>input</var> stream,
 runs these steps:
@@ -1603,7 +1606,6 @@ runs these steps:
 <p class=note>This is equivalent to the <a spec=webidl>convert a DOMString to a sequence of Unicode
 scalar values</a> algorithm from [[WEBIDL]], but allows for surrogate pairs that are split between
 strings.
-
 
 <p>The <dfn id=concept-te-encode-and-flush>encode and flush</dfn> algorithm, given a
 {{TextEncoder}} <var>enc</var>, runs these steps:
--- a/encoding.bs
+++ b/encoding.bs
@@ -1224,16 +1224,17 @@ constructor, when invoked, must run these steps:
  <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>ignore BOM flag</a> to <var>dec</var>'s
  <a for=TextDecoder>ignore BOM flag</a>.
 
- <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>decoder</a> to a new
- <a for=/>decoder</a> for <var>decForTransform</var>'s <a for=TextDecoder>encoding</a>, and set
- <var>decForTransform</var>'s <a for=TextDecoder>stream</a> to a new <a for=/>stream</a>.
+ <li>
+  <p>Set <var>decForTransform</var>'s <a for=TextDecoder>decoder</a> to a new <a for=/>decoder</a>
+  for <var>decForTransform</var>'s <a for=TextDecoder>encoding</a>, and set
+  <var>decForTransform</var>'s <a for=TextDecoder>stream</a> to a new <a for=/>stream</a>.
 
- <p class=note>For simplicity, <var>dec</var> and <var>decForTransform</var> have redundant
- members. However, the <a for=TextDecoder>BOM seen flag</a>, <a for=TextDecoder>do not flush
- flag</a>, and <a for=TextDecoder>transform</a> are unused in <var>decForTransform</var>, and
- <a for=TextDecoder>encoding</a>, <a for=TextDecoder>ignore BOM flag</a> and
- <a for=TextDecoder>error mode</a> are identical to <var>dec</var>. It is not necessary for
- implementations to duplicate these member fields.</p>
+  <p class=note>For simplicity, <var>dec</var> and <var>decForTransform</var> have redundant
+  members. However, the <a for=TextDecoder>BOM seen flag</a>, <a for=TextDecoder>do not flush
+  flag</a>, and <a for=TextDecoder>transform</a> are unused in <var>decForTransform</var>, and
+  <a for=TextDecoder>encoding</a>, <a for=TextDecoder>ignore BOM flag</a> and
+  <a for=TextDecoder>error mode</a> are identical to <var>dec</var>. It is not necessary for
+  implementations to duplicate these member fields.
 
  <li><p>Let <var>startAlgorithm</var> be an algorithm that takes no arguments and returns nothing.
 
@@ -1462,13 +1463,14 @@ constructor, when invoked, must run these steps:
 
  <li><p>Let <var>encForTransform</var> be a new {{TextEncoder}} object.
 
- <li><p>Set <var>encForTransform</var>'s <a for=TextEncoder>encoder</a> to <a>UTF-8</a>'s
- <a for=/>encoder</a>.
+ <li>
+  <p>Set <var>encForTransform</var>'s <a for=TextEncoder>encoder</a> to <a>UTF-8</a>'s
+  <a for=/>encoder</a>.
 
- <p class=note>For simplicity, <var>enc</var> and <var>encForTransform</var> have the same
- members. However, <a for=TextEncoder>transform</a> is not used by <var>encForTransform</var>
- and <a for=TextEncoder>pending high surrogate</a> is not used by <var>enc</var>. It is not
- necessary for implementations to store these unused member fields.</p>
+  <p class=note>For simplicity, <var>enc</var> and <var>encForTransform</var> have the same
+  members. However, <a for=TextEncoder>transform</a> is not used by <var>encForTransform</var> and
+  <a for=TextEncoder>pending high surrogate</a> is not used by <var>enc</var>. It is not necessary
+  for implementations to store these unused member fields.
 
  <li><p>Let <var>startAlgorithm</var> be an algorithm that takes no arguments and returns nothing.
 
@@ -1621,14 +1623,10 @@ between strings.
    <li><p>Let <var>controller</var> be <var>encForTransform</var>'s
    <a for=TextEncoder>transform</a>.\[[transformStreamController]].
 
-   <li><p>Let <var>token</var> be U+FFFD.
+   <li>
+    <p>Let <var>output</var> be the byte sequence 0xEF 0xBF 0xBD.
 
-   <li><p>Let <var>input</var> and <var>output</var> be new <a for=/>stream</a>s.
-
-   <li><p><a>Process</a> <var>token</var> for <a for=TextEncoder>encoder</a>, <var>input</var>,
-   <var>output</var>.
-
-   <li><p>Convert <var>output</var> into a byte sequence.
+    <p class=note>This is the replacement character U+FFFD encoded as UTF-8.
 
    <li><p>Let <var>chunk</var> be a {{Uint8Array}} object wrapping an {{ArrayBuffer}} containing
    <var>output</var>.
--- a/encoding.bs
+++ b/encoding.bs
@@ -1214,13 +1214,35 @@ constructor, when invoked, must run these steps:
  <li><p>If <var>options</var>'s <code>ignoreBOM</code> member is true, then set <var>dec</var>'s
  <a for=TextDecoder>ignore BOM flag</a>.
 
+ <li><p>Let <var>decForTransform</var> be a new {{TextDecoder}} object.
+
+ <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>encoding</a> to <var>encoding</var>.
+
+ <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>error mode</a> to <var>dec</var>'s
+ <a for=TextDecoder>error mode</a>.
+
+ <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>ignore BOM flag</a> to <var>dec</var>'s
+ <a for=TextDecoder>ignore BOM flag</a>.
+
+ <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>decoder</a> to a new <a
+ for=/>decoder</a> for <var>decForTransform</var>'s <a for=TextDecoder>encoding</a>, and set
+ <var>decForTransform</var>'s <a for=TextDecoder>stream</a> to a new <a for=/>stream</a>.
+
+ <p class=note>For simplicity, <var>dec</var> and <var>decForTransform</var> have redundant
+ members. However, the <a for=TextDecoder>BOM seen flag</a>, <a for=TextDecoder>do not flush
+ flag</a>, and <a for=TextDecoder>transform</a> are unused in <var>decForTransform</var>, and <a
+ for=TextDecoder>encoding</a>, <a for=TextDecoder>ignore BOM flag</a> and <a for=TextDecoder>error
+ mode</a> are identical to <var>dec</var>. It is not necessary for implementations to duplicate
+ these member fields.</p>
+
  <li><p>Let <var>startAlgorithm</var> be an algorithm that takes no arguments and returns nothing.
 
  <li><p>Let <var>transformAlgorithm</var> be an algorithm which takes a <var>chunk</var> argument
- and runs the <a>decode and enqueue a chunk</a> algorithm with <var>dec</var> and <var>chunk</var>.
+ and runs the <a>decode and enqueue a chunk</a> algorithm with <var>decForTransform</var> and
+ <var>chunk</var>.
 
  <li><p>Let <var>flushAlgorithm</var> be an algorithm which takes no arguments and runs the <a>flush
- and enqueue</a> algorithm with <var>dec</var>.
+ and enqueue</a> algorithm with <var>decForTransform</var>.
 
  <li><p>Let <var>transform</var> be the result of calling
  <a abstract-op>CreateTransformStream</a>(<var>startAlgorithm</var>, <var>transformAlgorithm</var>,
@@ -1250,19 +1272,6 @@ true if <a for=TextDecoder>ignore BOM flag</a> is set, and false otherwise.
 method, when invoked, must run these steps:
 
 <ol>
- <li><p>Let <var>readable</var> be <a for=TextDecoder>transform</a>.\[[readable]].
-
- <li><p>If <a abstract-op>IsReadableStreamLocked</a>(<var>readable</var>) is true, then throw a
- {{TypeError}} exception.
-
- <li><p>Let <var>writable</var> be <a for=TextDecoder>transform</a>.\[[writable]].
-
- <li><p>If <a abstract-op>IsWritableStreamLocked</a>(<var>writable</var>) is true, then throw a
- {{TypeError}} exception.
-
- <p class="note">These steps ensure that the state of the <a for=/>decoder</a> is not simultaneously
- modified by the Streams API and this method.
-
  <li><p>If the <a for=TextDecoder>do not flush flag</a> is unset, set <a for=TextDecoder>decoder</a>
  to a new <a for=TextDecoder>encoding</a>'s <a for=/>decoder</a>, set <a for=TextDecoder>stream</a>
  to a new <a for=/>stream</a>, and unset the <a for=TextDecoder>BOM seen flag</a>.
@@ -1317,7 +1326,7 @@ method, when invoked, must run these steps:
 <hr>
 
 <p>The <dfn id=concept-td-decode-and-enqueue>decode and enqueue a chunk</dfn> algorithm, given a
-{{TextDecoder}} <var>dec</var> and a <var>chunk</var>, runs these steps:
+{{TextDecoder}} <var>decForTransform</var> and a <var>chunk</var>, runs these steps:
 
 <ol>
  <li><p>If <a>Type</a>(<var>chunk</var>) is not Object, or <var>chunk</var> does not have an
@@ -1325,17 +1334,10 @@ method, when invoked, must run these steps:
  true, or <a abstract-op>IsSharedArrayBuffer</a>(<var>chunk</var>) is true, then return a new
  promise rejected with a {{TypeError}} exception.
 
- <li><p>If the <a for=TextDecoder>do not flush flag</a> for <var>dec</var> is unset, then set
- <var>dec's</var> <a for=TextDecoder>decoder</a> to a new <a for=/>decoder</a> for <var>dec</var>'s
- <a for=TextDecoder>encoding</a>, set <var>dec</var>'s <a for=TextDecoder>stream</a> to a new
- <a for=/>stream</a>, and unset <var>dec</var>'s <a for=TextDecoder>BOM seen flag</a>.
-
- <li><p>Set <var>dec</var>'s <a for=TextDecoder>do not flush flag</a>.
-
  <li><p><a>Push</a> a <a lt="get a copy of the buffer source">copy of</a> <var>chunk</var> to
- <var>dec</var>'s <a for=TextDecoder>stream</a>.
+ <var>decForTransform</var>'s <a for=TextDecoder>stream</a>.
 
- <li><p>Let <var>controller</var> be <var>dec</var>'s
+ <li><p>Let <var>controller</var> be <var>decForTransform</var>'s
  <a for=TextDecoder>transform</a>.\[[transformStreamController]].
 
  <li><p>Let <var>output</var> be a new <a for=/>stream</a>.
@@ -1344,7 +1346,7 @@ method, when invoked, must run these steps:
   <p>While true, run these steps:
 
   <ol>
-   <li><p>Let <var>token</var> be the result of <a>reading</a> from <var>dec</var>'s
+   <li><p>Let <var>token</var> be the result of <a>reading</a> from <var>decForTransform</var>'s
    <a for=TextDecoder>stream</a>.
 
    <li>
@@ -1360,8 +1362,9 @@ method, when invoked, must run these steps:
     </ol>
 
    <li><p>Let <var>result</var> be the result of <a>processing</a> <var>token</var> for
-   <var>dec</var>'s <a for=TextDecoder>decoder</a>, <var>dec</var>'s <a for=TextDecoder>stream</a>,
-   <var>output</var>, and <var>dec's</var> <a for=TextDecoder>error mode</a>.
+   <var>decForTransform</var>'s <a for=TextDecoder>decoder</a>, <var>decForTransform</var>'s <a
+   for=TextDecoder>stream</a>, <var>output</var>, and <var>decForTransform</var>'s <a
+   for=TextDecoder>error mode</a>.
 
    <li><p>If <var>result</var> is <a>error</a>, then return a new promise rejected with a
    {{TypeError}} exception.
@@ -1369,27 +1372,22 @@ method, when invoked, must run these steps:
 </ol>
 
 <p>The <dfn id=concept-td-flush-and-enqueue>flush and enqueue</dfn> algorithm, which handles the end
-of data from the input {{ReadableStream}}, given a {{TextDecoder}} <var>dec</var>, runs these steps:
+of data from the input {{ReadableStream}}, given a {{TextDecoder}} <var>decForTransform</var>, runs
+these steps:
 
 <ol>
- <li><p>If the <a for=TextDecoder>do not flush flag</a> for <var>dec</var> is unset, then set
- <var>dec's</var> <a for=TextDecoder>decoder</a> to a new <a for=/>decoder</a> for <var>dec</var>'s
- <a for=TextDecoder>encoding</a>, set <var>dec</var>'s <a for=TextDecoder>stream</a> to a new
- <a for=/>stream</a>, and unset <var>dec</var>'s <a for=TextDecoder>BOM seen flag</a>.
-
- <li><p>Unset <var>dec</var>'s <a for=TextDecoder>do not flush flag</a>.
-
  <li><p>Let <var>output</var> be a new <a for=/>stream</a>.
 
  <li><p>Let <var>result</var> be the result of <a>processing</a> <a>end-of-stream</a> for
- <var>dec</var>'s <a for=TextDecoder>decoder</a> and <var>dec</var>'s <a for=TextDecoder>stream</a>,
- <var>output</var>, and <var>dec</var>'s <a for=TextDecoder>error mode</a>.
+ <var>decForTransform</var>'s <a for=TextDecoder>decoder</a> and <var>decForTransform</var>'s <a
+ for=TextDecoder>stream</a>, <var>output</var>, and <var>decForTransform</var>'s <a
+ for=TextDecoder>error mode</a>.
 
  <li><p>If <var>result</var> is <a>finished</a>, then run these steps:
  <ol>
   <li><p>Let <var>outputChunk</var> be <var>output</var>, <a lt="serialize stream">serialized</a>.
 
-  <li><p>Let <var>controller</var> be <var>dec</var>'s
+  <li><p>Let <var>controller</var> be <var>decForTransform</var>'s
   <a for=TextDecoder>transform</a>.\[[transformStreamController]].
 
   <li><p>Call <a abstract-op>TransformStreamDefaultControllerEnqueue</a>(<var>controller</var>,
--- a/encoding.bs
+++ b/encoding.bs
@@ -1539,9 +1539,8 @@ must run these steps:
  with that exception.
 
  <p class=note>{{DOMString}} is used here so that a surrogate pair that is split between chunks can
- be correctly reassembled into the appropriate code point in the output. The behaviour is otherwise
- identical to {{USVString}}. In particular, lone surrogates will be replaced with U+FFFD so that the
- output is always well-formed UTF-8.
+ be reassembled into the appropriate scalar value. The behavior is otherwise
+ identical to {{USVString}}. In particular, lone surrogates will be replaced with U+FFFD.
 
  <li><p>Convert <var>input</var> to a <a for=/>stream</a>.
 
--- a/encoding.bs
+++ b/encoding.bs
@@ -1538,6 +1538,11 @@ must run these steps:
  <var>chunk</var> to a {{DOMString}}. If this throws an exception, then return a promise rejected
  with that exception.
 
+ <p class=note>{{DOMString}} is used here so that a surrogate pair that is split between chunks can
+ be correctly reassembled into the appropriate code point in the output. The behaviour is otherwise
+ identical to {{USVString}}. In particular, replacement characters will be used where necessary to
+ make the output valid UTF-8.
+
  <li><p>Convert <var>input</var> to a <a for=/>stream</a>.
 
  <li><p>Let <var>output</var> be a new <a for=/>stream</a>.
--- a/encoding.bs
+++ b/encoding.bs
@@ -1224,16 +1224,16 @@ constructor, when invoked, must run these steps:
  <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>ignore BOM flag</a> to <var>dec</var>'s
  <a for=TextDecoder>ignore BOM flag</a>.
 
- <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>decoder</a> to a new <a
- for=/>decoder</a> for <var>decForTransform</var>'s <a for=TextDecoder>encoding</a>, and set
+ <li><p>Set <var>decForTransform</var>'s <a for=TextDecoder>decoder</a> to a new
+ <a for=/>decoder</a> for <var>decForTransform</var>'s <a for=TextDecoder>encoding</a>, and set
  <var>decForTransform</var>'s <a for=TextDecoder>stream</a> to a new <a for=/>stream</a>.
 
  <p class=note>For simplicity, <var>dec</var> and <var>decForTransform</var> have redundant
  members. However, the <a for=TextDecoder>BOM seen flag</a>, <a for=TextDecoder>do not flush
- flag</a>, and <a for=TextDecoder>transform</a> are unused in <var>decForTransform</var>, and <a
- for=TextDecoder>encoding</a>, <a for=TextDecoder>ignore BOM flag</a> and <a for=TextDecoder>error
- mode</a> are identical to <var>dec</var>. It is not necessary for implementations to duplicate
- these member fields.</p>
+ flag</a>, and <a for=TextDecoder>transform</a> are unused in <var>decForTransform</var>, and
+ <a for=TextDecoder>encoding</a>, <a for=TextDecoder>ignore BOM flag</a> and
+ <a for=TextDecoder>error mode</a> are identical to <var>dec</var>. It is not necessary for
+ implementations to duplicate these member fields.</p>
 
  <li><p>Let <var>startAlgorithm</var> be an algorithm that takes no arguments and returns nothing.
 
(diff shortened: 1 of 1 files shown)


inexorabletash:
Yay! Comments in random order:

Since it's a common enough request, would it be possible to include an example in the spec itself showing how to use streams to encode into an existing ArrayBuffer?

Can the tests be upgraded to use `async`/`await` ?

In the tests, stream-properties.html should probably just be an idlharness test. (I don't know how well that plays with p(r)ollyfills, though.)

What's the behavior for split surrogate pairs when encoding? We basically pretend that's not a problem in the non-streaming API, by making the input is a USVString and thus not needing a `{stream}` flag. In other words, we assume that the input is always a complete text, not just a randomly partitioned sequence of code units. In this patch, the text reads _Let input be the result of converting chunk to a USVString...._ which sounds like it might introduce problems when chunk sizes cross surrogate boundaries. This worries me. Also, needs test cases.

ricea:
> Since it's a common enough request, would it be possible to include an example in the spec itself showing how to use streams to encode into an existing ArrayBuffer?

Sadly, bring-your-own-buffer style usage isn't supported by TransformStream yet. There's an open question about what to do when there isn't enough space in the supplied ArrayBuffer to fit the transformed input.

Since this is basically an optimisation feature, I don't plan to support it in the first release. I would rather take my time to get it right.

> Can the tests be upgraded to use async/await ?

Certainly. I'm never quite sure what version of ES to target in tests. For the Streams Standard tests I've been avoiding introducing async/await, but I'm happy to use them here.

> In the tests, stream-properties.html should probably just be an idlharness test. (I don't know how well that plays with p(r)ollyfills, though.)

I haven't tried it. I will give it a go.

> What's the behavior for split surrogate pairs when encoding? We basically pretend that's not a problem in the non-streaming API, by making the input is a USVString and thus not needing a {stream} flag. In other words, we assume that the input is always a complete text, not just a randomly partitioned sequence of code units. In this patch, the text reads Let input be the result of converting chunk to a USVString.... which sounds like it might introduce problems when chunk sizes cross surrogate boundaries. This worries me. Also, needs test cases.

Yes, it's broken. I've been planning to write a test for this at some point. What I expect to see is if you encode "😻" in a single chunk, it works as intended. If you split it into two chunks, you get two chunks as output each containing U+FFFD encoded as UTF-8.

If there is consensus that this should be fixed, I think it can be done by converting to DOMString rather than USVString and special-casing the final code unit when it is the first half of a surrogate pair.

annevk:
I think a problem here is that if a script running in a document invokes the constructor of either `TextDecoder` or `TextEncoder` in an `<iframe>` element the tear-off transform-related objects will have the associated global of the document (the current global/realm) and not the `<iframe>`'s document.

See also https://www.w3.org/Bugs/Public/show_bug.cgi?id=24652 and https://w3c-test.org/WebIDL/current-realm.html.

Note that if you change it to accept a DOMString you also need to do conversions from [code unit](https://infra.spec.whatwg.org/#code-unit) to [scalar value](https://infra.spec.whatwg.org/#scalar-value) (you probably meant that; stating it here so I don't forget to verify it).

(Tests can use JavaScript features that ship in the latest release of all browsers. At least, that's my rule of thumb.)

domenic:
I think we should indeed make split surrogates work. It seems like a big value proposition for having this built-in and shared, that people don't have to figure that out by themselves.

> I think a problem here is that if a script running in a document invokes the constructor of either TextDecoder or TextEncoder in an `<iframe>` element the tear-off transform-related objects will have the associated global of the document (the current global/realm) and not the `<iframe>`'s document.

I don't think that is true. You're talking about `new childIframe.contentWindow.TextEncoder()` The current realm inside the TextEncoder constructor is that of the constructor itself, i.e. of the iframe, so CreateTransformStream will still be in that realm, and the `TransformStream` objects will be from the same realm as the TextEncoder object constructed.

annevk:
Ah, great, thanks for confirming. Let's make sure to test that as part of the test I referenced.

ricea:
@inexorabletash https://github.com/GoogleChromeLabs/text-encode-transform-prollyfill/pull/6 updates the test to use async / await (much more readable) and adds idlharness.html. The prollyfill fails a couple of the idlharness tests because it (intentionally) doesn't do brand checks. Apart from that, everything looks good.

ricea:
I have been writing tests for surrogate pairs split between chunks and in the process realised there's a problem.

We're now explicitly taking a sequence of code units and normalising it to Unicode. Which means we're doing error handling. Which means the developer should probably be able to specify what type of error handling should be done. Which means adding a fatal option to the constructor. Which means that encode() should logically also support it. Which means that encode() also needs to be changed to take a DOMString. Which is a much more disruptive change than this one.

My preference is to land this first, and then worry about providing control over error behaviour and changing encode() later, since that is a much riskier change.

annevk:
> Which means we're doing error handling.

Well, but not any different than IDL always did, right? We're just changing the layering a bit. I'm not necessarily convinced String -> USVString conversion needs to be subject to fatal the same way that bytes <> scalar values needs to be. Note that the reason we have it in the latter case is primarily XML, which is the only format I know of that demands such handling. We exposed it to the API as a nicety mostly.

ricea:
> Well, but not any different than IDL always did, right? We're just changing the layering a bit. I'm not necessarily convinced String -> USVString conversion needs to be subject to fatal the same way that bytes <> scalar values needs to be. Note that the reason we have it in the latter case is primarily XML, which is the only format I know of that demands such handling. We exposed it to the API as a nicety mostly.

That works for me too.

ricea:
I have written [tests for split surrogates](https://github.com/GoogleChromeLabs/text-encode-transform-prollyfill/pull/8) but I haven't written the standard language yet.

I am working on tests for ensuring that objects are all created in the correct realms. This is taking a while.

I probably won't have the standard language updated for split surrogates until the new year.

annevk:
@vyv03354 are you willing to review this as well? Thanks!

ricea:
I have updated the standard language for split surrogates. Sorry for the delay. I tried a few approaches and adding the "convert code unit to scalar value" algorithm seemed to work best. PTAL.

encoding.bs
```diff
@@ -28,6 +28,24 @@ Translate IDs: dictdef-textdecoderoptions textdecoderoptions,dictdef-textdecodeo
...
+</pre>
+
+<pre class=anchors>
+spec: streams; urlPrefix: https://streams.spec.whatwg.org/
+    type:dfn; text:chunk; url:#chunk
+    type:dfn; text:readable stream; url:#readable-stream
+    type:dfn; text:writable stream; url:#writable-stream
+    type:abstract-op; text:CreateTransformStream; url: #create-transform-stream
+    type:abstract-op; text:TransformStreamDefaultControllerEnqueue; url: #transform-stream-default-controller-enqueue
+    type:interface; text:TransformStream; url: #ts-class
```
annevk: What's not exported yet in streams should be exported so there's no need to list it here.

encoding.bs
```diff
@@ -28,6 +28,24 @@ Translate IDs: dictdef-textdecoderoptions textdecoderoptions,dictdef-textdecodeo
...
+spec: ECMASCRIPT; urlPrefix: https://tc39.github.io/ecma262/
+    text: type; url: #sec-ecmascript-data-types-and-values; type: dfn
+    text: IsDetachedBuffer; url: #sec-isdetachedbuffer; type: abstract-op
+    text: IsSharedArrayBuffer; url: #sec-issharedarraybuffer; type: abstract-op
+    text: internal slot; url: #sec-object-internal-methods-and-internal-slots; type: dfn
```
annevk: We generally don't cross-reference "internal slot".

encoding.bs
```diff
@@ -1186,10 +1246,31 @@ if <a for=TextDecoder>error mode</a> is "<code>fatal</code>", and false otherwis
 <p>The <dfn attribute for=TextDecoder><code>ignoreBOM</code></dfn> attribute's getter must return
 true if <a for=TextDecoder>ignore BOM flag</a> is set, and false otherwise.
 
+<p>The <dfn attribute for=TextDecoder><code>readable</code></dfn> attribute's getter must return the
+contents of <a for=TextDecoder>transform</a>'s \[[readable]] <a>internal slot</a>.
```
annevk: I think rather than "the contents of x" we usually say "x's value". @domenic?

encoding.bs
```diff
@@ -1186,10 +1246,31 @@ if <a for=TextDecoder>error mode</a> is "<code>fatal</code>", and false otherwis
...
+
 <p>The <dfn method for=TextDecoder><code>decode(<var>input</var>, <var>options</var>)</code></dfn>
 method, when invoked, must run these steps:
 
 <ol>
+ <li><p>Let <var>readable</var> be <a for=TextDecoder>transform</a>'s \[[readable]] <a>internal
+ slot</a>.
+
+ <li><p>If <a abstract-op>IsReadableStreamLocked</a>(<var>readable</var>) is true, throw a
+ {{TypeError}} exception.
```
annevk: then throw* (xref throw?)

encoding.bs
```diff
@@ -1241,6 +1322,88 @@ method, when invoked, must run these steps:
   </ol>
 </ol>
 
+<p>The <dfn id=concept-td-decode-and-enqueue>decode and enqueue a chunk</dfn> algorithm, given a
+{{TextDecoder}} <var>dec</var> and a <var>chunk</var>, runs these steps:
+
+<ol>
+ <li><p>If the <a>type</a> of of <var>chunk</var> is not Object, or <var>chunk</var> does not have
```
annevk: Shouldn't this be noted as Type(chunk)?

encoding.bs
```diff
@@ -1241,6 +1322,88 @@ method, when invoked, must run these steps:
...
+ an \[[ArrayBufferData]] <a>internal slot</a>, or <a
+ abstract-op>IsDetachedBuffer</a>(<var>chunk</var>) is true, or <a
+ abstract-op>IsSharedArrayBuffer</a>(<var>chunk</var>) is true then return <a>a promise rejected
```
annevk: Comma before then. Also I'm not sure I want to reference the promises guide as it really needs to be folded into IDL and has a bunch of things that confuse folks. Can we just say "a new promise rejected with a TypeError exception"?

encoding.bs
```diff
@@ -1241,6 +1322,88 @@ method, when invoked, must run these steps:
...
+ with</a> a {{TypeError}}.
+
+ <li><p>If the <a for=TextDecoder>do not flush flag</a> for <var>dec</var> is unset, set
```
annevk: then set*

encoding.bs
```diff
@@ -1241,6 +1322,88 @@ method, when invoked, must run these steps:
...
+ <li><p><a>Push</a> a <a lt="get a copy of the buffer source">copy of</a> <var>chunk</var> to
+ <var>dec</var>'s <a for=TextDecoder>stream</a>.
+
+ <li><p>Let <var>controller</var> be <var>dec</var>'s <a for=TextDecoder>transform</a>'s
+ \[[transformStreamController]] <a>internal slot</a>.
+
+ <li><p>Let <var>output</var> be a new <a for=/>stream</a>.
+
+ <li>
+  <p>While true, run these substeps:
```
annevk: steps* we try to avoid "substeps" and friends these days.

encoding.bs
```diff
@@ -1241,6 +1322,88 @@ method, when invoked, must run these steps:
...
+  <li><p>Let <var>controller</var> be <var>dec</var>'s <a for=TextDecoder>transform</a>'s
+  \[[transformStreamController]] <a>internal slot</a>.
+
+  <li><p>Call <a abstract-op>TransformStreamDefaultControllerEnqueue</a>(<var>controller</var>, <var>outputChunk</var>).
+
+  <li><p>Return <a>a promise resolved with</a> undefined.
+ </ol>
+
+ <li><p>Otherwise, return <a>a promise rejected with</a> a {{TypeError}}.
+</ol>
```
annevk: It seems like you ate a newline here.

encoding.bs
```diff
@@ -1267,6 +1434,23 @@ requires buffering of scalar values.
...
+  <a>UTF-8</a>'s <a for=/>encoder</a> on the chunks written to {{TextEncoder/writable}}.
+
+ <dt><code><var>encoder</var> . <a attribute for=TextEncoder>writable</a></code>
+ <dd>
+  <p>Returns a <a>writable stream</a> which accepts string chunks and runs them through
+  <a>UTF-8</a>'s <a for=/>encoder</a> before making them available to {{TextEncoder/readable}}.
+
+  <p>Typically this will be used via the {{ReadableStream/pipeThrough()}} method on a {{ReadableStream}} source.
+
+<pre class=example id=example-textencode-writable>
```
annevk: `<pre` also needs two spaces in front of it.

annevk: Also, this should probably use `<code>` too, just like the other examples.

hsivonen:
> Since it's a common enough request, would it be possible to include an example in the spec itself showing how to use streams to encode into an existing ArrayBuffer?

I'd like to reiterate the concerns about interoperability hazards associated with existing ArrayBuffer that I mentioned in [the issue dedicated to that request](https://github.com/whatwg/encoding/issues/69).

> I think we should indeed make split surrogates work.

While JS strings need to be able to exist in invalid-UTF-16 states when a JS program constructs a valid UTF-16 string, making unpaired surrogate preservation in crossing the WebIDL boundary into an explicit feature is a novel thing for the Web Platform and one that makes me uneasy.

So far, we have USVString, which explicitly replaces unpaired surrogates and various DOMString APIs that in UTF-16-based Web engines pass unpaired surrogates through, but that's mostly incidental and Web developers don't have good reasons to pass unpaired surrogates through, so not preserving unpaired surrogates appears to be feasible in terms of Web compat.

In Gecko, we were able to make CSSOM effectively change from DOMString to USVString without breaking the Web. If I've read the Servo source comments correctly, Servo panics if a DOMString contains unpaired surrogates and the absence of such crash reports suggests that the Web doesn't actually depend on unpaired surrogates crossing the WebIDL boundary (to the extent Servo has enough DOM features to be exercised with real Web content).

While of all the Web Platform APIs this might be the one where letting unpaired surrogates cross the WebIDL boundary might make sense as an explicit feature, I think we should consider a bit more carefully if that's necessary and if it's feasible to require textual streams not to split surrogate pairs across chunk boundaries.

encoding.bs
annevk: Ah yeah, indeed.

ricea:
@hsivonen Consider the following transform stream:

```javascript
const rechunker = new TransformStream({
  transform(chunk, controller) {
    for (let start = 0; start < chunk.length; start += 1024) {
      const end = Math.min(start + 1024, chunk.length);
      controller.enqueue(chunk.substring(start, end));
    }
  }
});
```
This takes a stream of strings and splits up any chunk that is longer than 1024 code units.

Here are the questions I ask myself:
1. _Is this a useful and sensible thing to want to do?_ I believe so. If the stream you're processing may have arbitrarily large chunks, you're forwarding it to some expensive operation, and you want to avoid jank, I think the correct approach is to include something like this in your pipe.
2. _It is reasonable for a developer to expect this to work and not corrupt the input?_ I think so. 
   1. Corruption wouldn't even happen unless you streamed it to `TextEncoder`, and that could be several steps away down the pipe.
   2. It just applies the Javascript built-in `substring()` function, and nothing on 
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring says anything about data corruption.
   3. It will be a long time before any problems are noticed.
3. _Is this implementation correct?_ Maybe? Maybe not? I don't know.

hsivonen:
I'm not fully familiar with Streams. Where do the string chunks in the rechunker example cross the WebIDL boundary (if anywhere)? The controller seems to be expressed in ECMA spec style and not in WebIDL style. Is there really no WebIDL boundary here and the prose sentence "Let input be the result of converting chunk to a DOMString." defines the boundary that is usually expressed in WebIDL?

ricea:
> I'm not fully familiar with Streams. Where do the string chunks in the rechunker example cross the WebIDL boundary (if anywhere)?

Sorry for not explaining clearly. The chunks don't cross the WebIDL boundary within the rechunker example. It's only if you then pipe the stream to TextEncoder or some other stream defined in a WebIDL-based standard that the chunks will cross the WebIDL boundary.

>  The controller seems to be expressed in ECMA spec style and not in WebIDL style. Is there really no WebIDL boundary here and the prose sentence "Let input be the result of converting chunk to a DOMString." defines the boundary that is usually expressed in WebIDL?

I think that is correct. My intent is for "converting chunk to a DOMString" to imply the same processing that would normally happen upon calling a WebIDL method. Is this something that needs to be made more explicit? The same applies to the first step of the "decode and enqueue a chunk" algorithm.

annevk:
@ricea if transform streams gain a "bring your own buffer" capability does that need explicit changes to this API or would we run into https://github.com/whatwg/encoding/issues/69#issuecomment-349946317 as a side effect of that capability being added? In case it's the latter, we'll need to address @hsivonen's concerns somehow now, I think.

ricea:
@annevk It won't be an automatic upgrade. There will be a different Create operation to call, similar to [CreateReadableByteStream](https://streams.spec.whatwg.org/#create-readable-byte-stream). While this is a bit regrettable, it would be bad if a change in the Streams Standard suddenly exposed internals of Encoding and other standards.

encoding.bs
```diff
@@ -28,6 +28,18 @@ Translate IDs: dictdef-textdecoderoptions textdecoderoptions,dictdef-textdecodeo
 spec:infra; type:dfn;
     text:code point
     text:ascii case-insensitive
+spec:streams;
+    type:interface; text:ReadableStream
+    type:dfn; text:chunk
+    type:dfn; text:readable stream
+    type:dfn; text:writable stream
```
annevk: Why are these needed? Do you get some conflict that cannot be resolved?

encoding.bs
```diff
@@ -1066,15 +1078,18 @@ interface TextDecoder {
...
 
 <p>A {{TextDecoder}} object has an associated <dfn for=TextDecoder>encoding</dfn>,
 <dfn for=TextDecoder>decoder</dfn>, <dfn for=TextDecoder>stream</dfn>,
 <dfn for=TextDecoder>ignore BOM flag</dfn> (initially unset),
 <dfn for=TextDecoder>BOM seen flag</dfn> (initially unset),
-<dfn for=TextDecoder>error mode</dfn> (initially "<code>replacement</code>"), and
-<dfn for=TextDecoder>do not flush flag</dfn> (initially unset).
+<dfn for=TextDecoder>error mode</dfn> (initially "<code>replacement</code>"),
+<dfn for=TextDecoder>do not flush flag</dfn> (initially unset), and
+<dfn for=TextDecoder>transform</dfn> (a {{TransformStream}} object).
```
annevk: It seems best to not say anything here between parenthesis as that's only used for initial values currently (which is inconsistent with other specifications, I realize). We could do a follow-up where we state the type for all of these and then put the initial value elsewhere.

encoding.bs
```diff
@@ -1135,6 +1150,31 @@ control.
  <dt><code><var>decoder</var> . <a attribute for=TextDecoder>ignoreBOM</a></code>
  <dd><p>Returns true if <a for=TextDecoder>ignore BOM flag</a> is set, and false otherwise.
 
+ <dt><code><var>decoder</var> . <a attribute for=TextDecoder>readable</a></code>
+ <dd>
+  <p>Returns a <a>readable stream</a> whose <a>chunks</a> are strings resulting from running <a
```
annevk: Please don't wrap inline/phrasing elements in any manner.

encoding.bs
```diff
@@ -1135,6 +1150,31 @@ control.
...
+  for=TextDecoder>encoding</a>'s <a for=/>decoder</a> on the chunks written to
+  {{TextDecoder/writable}}.
+
+ <dt><code><var>decoder</var> . <a attribute for=TextDecoder>writable</a></code>
+ <dd>
+  <p>Returns a <a>writable stream</a> which accepts {{BufferSource}} chunks and runs them through <a
```
annevk: Same here.

encoding.bs
```diff
@@ -1135,6 +1150,31 @@ control.
...
+  {{ReadableStream}} source.
+
+  <pre class=example id=example-textdecode-writable><code class=lang-javascript>
+var decoder = new TextDecoder(encoding);
+byteReadable
+  .pipeThrough(decoder)
+  .pipeTo(textWritable);</code></pre>
+
+  <p>If the <a for=TextDecoder>error mode</a> is "<code>fatal</code>" and
+  <a for=TextDecoder>encoding</a>'s <a for=/>decoder</a> returns <a>error</a>, both {{TextDecoder/readable}}
```
annevk: This exceeds 100 columns.

encoding.bs
```diff
@@ -1174,6 +1214,20 @@ constructor, when invoked, must run these steps:
  <li><p>If <var>options</var>'s <code>ignoreBOM</code> member is true, then set <var>dec</var>'s
  <a for=TextDecoder>ignore BOM flag</a>.
 
+ <li><p>Let <var>startAlgorithm</var> be an algorithm that takes no arguments and returns nothing.
+
+ <li><p>Let <var>transformAlgorithm</var> be an algorithm which takes a <var>chunk</var> argument
+ and runs the <a>decode and enqueue a chunk</a> algorithm with <var>dec</var> and <var>chunk</var>.
+
+ <li><p>Let <var>flushAlgorithm</var> be an algorithm which takes no arguments and runs the <a>flush
```
annevk: No wrapping inside phrasing elements.

encoding.bs
```diff
@@ -1174,6 +1214,20 @@ constructor, when invoked, must run these steps:
...
+ and enqueue</a> algorithm with <var>dec</var>.
+
+ <li><p>Let <var>transform</var> be the result of calling <a
```
annevk: No wrapping.

encoding.bs
```diff
@@ -1186,15 +1240,34 @@ if <a for=TextDecoder>error mode</a> is "<code>fatal</code>", and false otherwis
...
+
+ <li><p>If <a abstract-op>IsReadableStreamLocked</a>(<var>readable</var>) is true, then throw a
+ {{TypeError}} exception.
+
+ <li><p>Let <var>writable</var> be <a for=TextDecoder>transform</a>.\[[writable]].
+
+ <li><p>If <a abstract-op>IsWritableStreamLocked</a>(<var>writable</var>) is true, then throw a
+ {{TypeError}} exception.
+
+ <p class="note">These steps ensure that the state of the <a for=/>decoder</a> is not simultaneously modified by
```
annevk: Longer than a 100 columns.

encoding.bs
```diff
@@ -1241,6 +1314,88 @@ method, when invoked, must run these steps:
...
+  <ol>
+   <li><p>Let <var>token</var> be the result of <a>reading</a> from <var>dec</var>'s <a
+   for=TextDecoder>stream</a>.
+
+   <li>
+    <p>If <var>token</var> is <a>end-of-stream</a>, run these steps:
+    <ol>
+     <li><p>Let <var>outputChunk</var> be <var>output</var>, <a lt="serialize stream">serialized</a>.
+
+     <li><p>Call <a abstract-op>TransformStreamDefaultControllerEnqueue</a>(<var>controller</var>, <var>outputChunk</var>).
```
annevk: Here you can wrap before `<var>outputChunk</var>).`

encoding.bs
```diff
@@ -1241,6 +1314,88 @@ method, when invoked, must run these steps:
...
+
+     <li><p>Return a new promise resolved with undefined.
+    </ol>
+
+   <li><p>Let <var>result</var> be the result of <a>processing</a> <var>token</var> for
+   <var>dec</var>'s <a for=TextDecoder>decoder</a>, <var>dec</var>'s <a for=TextDecoder>stream</a>,
+   <var>output</var>, and <var>dec's</var> <a for=TextDecoder>error mode</a>.
+
+   <li><p>If <var>result</var> is <a>error</a>, return a new promise rejected with a
```
annevk: then return

encoding.bs
```diff
@@ -1241,6 +1314,88 @@ method, when invoked, must run these steps:
...
+   {{TypeError}} exception.
+  </ol>
+</ol>
+
+<p>The <dfn id=concept-td-flush-and-enqueue>flush and enqueue</dfn> algorithm, which handles the end
+of data from the input {{ReadableStream}}, given a {{TextDecoder}} <var>dec</var>, runs these steps:
+
+<ol>
+ <li><p>If the <a for=TextDecoder>do not flush flag</a> for <var>dec</var> is unset, set
```
annevk: then set*

encoding.bs
```diff
@@ -1241,6 +1314,88 @@ method, when invoked, must run these steps:
...
+ <var>output</var>, and <var>dec</var>'s <a for=TextDecoder>error mode</a>.
+
+ <li><p>If <var>result</var> is <a>finished</a>, run these steps:
+ <ol>
+  <li><p>Let <var>outputChunk</var> be <var>output</var>, <a lt="serialize stream">serialized</a>.
+
+  <li><p>Let <var>controller</var> be <var>dec</var>'s
+  <a for=TextDecoder>transform</a>.\[[transformStreamController]].
+
+  <li><p>Call <a abstract-op>TransformStreamDefaultControllerEnqueue</a>(<var>controller</var>, <var>outputChunk</var>).
```
annevk: Needs wrapping.

encoding.bs
```diff
@@ -1249,10 +1404,14 @@ method, when invoked, must run these steps:
...
   readonly attribute DOMString encoding;
+  [SameObject] readonly attribute ReadableStream readable;
+  [SameObject] readonly attribute WritableStream writable;
   [NewObject] Uint8Array encode(optional USVString input = "");
 };</pre>
 
-<p>A {{TextEncoder}} object has an associated <dfn for=TextEncoder>encoder</dfn>.
+<p>A {{TextEncoder}} object has an associated <dfn for=TextEncoder>encoder</dfn>, <dfn
+for=TextEncoder>transform</dfn> (a {{TransformStream}} object) and <dfn for=TextEncoder>pending high
+surrogate</dfn> (initially unset).
```
annevk: No wrapping inside phrasing elements. This also has the same problem where types and initial values are mixed. I recommend omitting types for now and filing a follow-up issue.

encoding.bs
```diff
@@ -1267,6 +1426,24 @@ requires buffering of scalar values.
...
+  <a>UTF-8</a>'s <a for=/>encoder</a> on the chunks written to {{TextEncoder/writable}}.
+
+ <dt><code><var>encoder</var> . <a attribute for=TextEncoder>writable</a></code>
+ <dd>
+  <p>Returns a <a>writable stream</a> which accepts string chunks and runs them through
+  <a>UTF-8</a>'s <a for=/>encoder</a> before making them available to {{TextEncoder/readable}}.
+
+  <p>Typically this will be used via the {{ReadableStream/pipeThrough()}} method on a {{ReadableStream}} source.
+
+
```
annevk: One newline can go here.

encoding.bs
```diff
@@ -1279,16 +1456,49 @@ constructor, when invoked, must run these steps:
...
+ and runs the <a>encode and enqueue a chunk</a> algorithm with <var>enc</var> and <var>chunk</var>.
+
+ <li><p>Let <var>flushAlgorithm</var> be an algorithm which runs the <a>encode and flush</a>
+ algorithm with <var>enc</var>.
+
+ <li><p>Let <var>transform</var> be the result of calling <a
+ abstract-op>CreateTransformStream</a>(<var>startAlgorithm</var>, <var>transformAlgorithm</var>,
+ <var>flushAlgorithm</var>).
+
+ <li><p>Set <var>enc</var>'s <a for=TextEncoder>transform</a> to <var>transform</var>
```
annevk: Needs a trailing dot.

encoding.bs
```diff
@@ -1279,16 +1456,49 @@ constructor, when invoked, must run these steps:
...
+
  <li><p>Return <var>enc</var>.
 </ol>
 
 <p>The <dfn attribute for=TextEncoder><code>encoding</code></dfn> attribute's getter must return
 "<code>utf-8</code>".
 
+<p>The <dfn attribute for=TextEncoder><code>readable</code></dfn> attribute's getter must return <a
```
annevk: No wrapping phrasing elements.

encoding.bs
```diff
@@ -1279,16 +1456,49 @@ constructor, when invoked, must run these steps:
...
+for=TextEncoder>transform</a>.\[[readable]].
+
+<p>The <dfn attribute for=TextEncoder><code>writable</code></dfn> attribute's getter must return <a
```
annevk: Same.

encoding.bs
```diff
@@ -1279,16 +1456,49 @@ constructor, when invoked, must run these steps:
...
+
+ <li><p>If <a abstract-op>IsReadableStreamLocked</a>(<var>readable</var>) is true, then throw a
+ {{TypeError}} exception.
+
+ <li><p>Let <var>writable</var> be <a for=TextEncoder>transform</a>.\[[writable]].
+
+ <li><p>If <a abstract-op>IsWritableStreamLocked</a>(<var>writable</var>) is true, then throw a
+ {{TypeError}} exception.
+
+ <p class="note">These steps are for consistent behaviour with the {{TextDecoder}} <a method
```
annevk: behavior*

Added to https://wiki.whatwg.org/wiki/Style

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
   </ol>
 </ol>
 
+<p>The <dfn id=concept-te-encode-and-enqueue>encode and enqueue a chunk</dfn> algorithm, given a
+{{TextEncoder}} <var>enc</var> and <var>chunk</var>, runs these steps:
+
+<ol>
+ <li><p>Let <var>input</var> be the result of <a lt="converted to an IDL value">converting</a>
+ <var>chunk</var> to a {{DOMString}}. If this throws an exception, return a promise rejected with
```
annevk: then return

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+ that exception.
+
+ <li><p>Convert <var>input</var> to a <a for=/>stream</a>.
+
+ <li><p>Let <var>output</var> be a new <a for=/>stream</a>.
+
+ <li><p>Let <var>encoder</var> be <a>UTF-8</a>'s <a for=/>encoder</a>.
+
+ <li><p>Let <var>controller</var> be <var>enc</var>'s <a
```
annevk: No wrapping inside phrasing elements.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+ for=TextEncoder>transform</a>.\[[transformStreamController]].
+
+ <li><p>While true, run these steps:
+
+ <ol>
+  <li><p>Let <var>token</var> be the result of <a>reading</a> from <var>input</var>.
+
+  <li><p>If <var>token</var> is <a>end-of-stream</a>, then run these steps:
```
annevk: Since this `<li>` has multiple children the `<p>` needs to go on a newline indented 1 (and the same for the following `<ol>`).

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+
+   <li><p>Return a new promise resolved with undefined.
+  </ol>
+
+  <li><p>Let <var>result</var> be the result of executing the <a>convert code unit to scalar
+  value</a> algorithm with <var>enc</var>, <var>token</var> and <var>input</var>.
+
+  <li><p>If <var>result</var> is not <a>continue</a>, <a>process</a> <var>result</var> for
+  <var>encoder</var>, <var>input</var>, <var>output</var>.
+
```
annevk: This newline can go.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+ <li><p>Convert <var>input</var> to a <a for=/>stream</a>.
+
+ <li><p>Let <var>output</var> be a new <a for=/>stream</a>.
+
+ <li><p>Let <var>encoder</var> be <a>UTF-8</a>'s <a for=/>encoder</a>.
+
+ <li><p>Let <var>controller</var> be <var>enc</var>'s <a
+ for=TextEncoder>transform</a>.\[[transformStreamController]].
+
+ <li><p>While true, run these steps:
```
annevk: Since this `<li>` has multiple children the `<p>` needs to go on a newline indented 1 (and the same for the following `<ol>`).

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+  <li><p>Let <var>result</var> be the result of executing the <a>convert code unit to scalar
+  value</a> algorithm with <var>enc</var>, <var>token</var> and <var>input</var>.
+
+  <li><p>If <var>result</var> is not <a>continue</a>, <a>process</a> <var>result</var> for
+  <var>encoder</var>, <var>input</var>, <var>output</var>.
+
+ </ol>
+</ol>
+
+
```
annevk: This newline can go. Or if you think that they should be separated even visually use `<hr>`.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+<p>The <dfn id=concept-te-convert-code-unit-to-scalar-value>convert code unit to scalar value</dfn>
+algorithm, given a {{TextEncoder}} <var>enc</var>, <var>token</var> and <var>input</var> stream,
+runs these steps:
+
+<ol>
+ <li><p>If <var>enc</var>'s <a>pending high surrogate</a> is set, run these steps:
```
annevk: `<li>` with multiple children again.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+  <li><p>If <var>token</var> is in the range U+DC00 to U+DFFF, inclusive, return a code point whose
+  value is 0x10000 + ((<var>high surrogate</var> &minus; 0xD800) &lt;&lt; 10) + (<var>token</var>
+  &minus; 0xDC00).
+
+  <li><p><a>Prepend</a> <var>token</var> to <var>input</var>.
+
+  <li><p>Return U+FFFD.
+ </ol>
+
+ <li><p>If <var>token</var> is in the range U+D800 to U+DBFF, inclusive, set <a>pending high
```
annevk: No newline inside a phrasing element.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+ <li><p>If <var>token</var> is in the range U+DC00 to U+DFFF, inclusive, return U+FFFD.
+
+ <li><p>Return <var>token</var>.
+</ol>
+
+<p class=note>This is equivalent to the <a spec=webidl>convert a DOMString to a sequence of Unicode
+scalar values</a> algorithm from [[WEBIDL]], but allows for surrogate pairs that are split between
+strings.
+
+
```
annevk: No second newline here (or potentially `<hr>`).

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+<p>The <dfn id=concept-te-encode-and-flush>encode and flush</dfn> algorithm, given a
+{{TextEncoder}} <var>enc</var>, runs these steps:
+
+<ol>
+ <li><p>If <var>enc</var>'s <a>pending high surrogate</a> is set, run these steps:
```
annevk: `<li>` with multiple children again.

encoding.bs
```diff
@@ -1241,6 +1314,88 @@ method, when invoked, must run these steps:
   </ol>
 </ol>
```
annevk: Having an `<hr>` here would be good I think, to separate it a bit from the member definitions.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+ surrogate</a> to <var>token</var> and return <a>continue</a>.
+
+ <li><p>If <var>token</var> is in the range U+DC00 to U+DFFF, inclusive, return U+FFFD.
+
+ <li><p>Return <var>token</var>.
+</ol>
+
+<p class=note>This is equivalent to the <a spec=webidl>convert a DOMString to a sequence of Unicode
+scalar values</a> algorithm from [[WEBIDL]], but allows for surrogate pairs that are split between
+strings.
```
annevk: I was thinking, it would be better to reference https://infra.spec.whatwg.org/#javascript-string-convert here since eventually we'll update IDL to use that too and then this would need to be updated.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
...
+algorithm, given a {{TextEncoder}} <var>enc</var>, <var>token</var> and <var>input</var> stream,
+runs these steps:
+
+<ol>
+ <li><p>If <var>enc</var>'s <a>pending high surrogate</a> is set, run these steps:
+
+ <ol>
+  <li><p>Let <var>high surrogate</var> be <var>enc</var>'s <a>pending high surrogate</a>.
+
+  <li><p>Unset <var>enc</var>'s <a>pending high surrogate</a>.
```
annevk: I think it would be slightly clearer if we instead made pending high surrogate's type "null or a code unit" instead of the current mixture of flag and code unit. So check for non-null and set it to null when done.

encoding.bs
```diff
@@ -1314,6 +1524,113 @@ must run these steps:
   </ol>
 </ol>
```
annevk: An `<hr>` would be good here too.

hsivonen:
> I think that is correct. My intent is for "converting chunk to a DOMString" to imply the same processing that would normally happen upon calling a WebIDL method. Is this something that needs to be made more explicit?

Please add a very conspicuous note explaining that unlike in cases where the JS string to DOMString boundary passes through data that doesn't conform to Unicode as a performance optimization / legacy artifact of failure to enforce Unicode conformance, in _this_ case letting code unit sequences that do not conform to Unicode travel across the boundary is an explicit feature with the intent of allowing the sequence of code units to  become conforming once concatenated on the other side of the JS string / DOMString boundary.

encoding.bs
ricea: "ReadableStream" is also exported by Fetch, and so needs to be disambiguated. I have modified the Streams Standard to export "chunk", "readable stream" and "writable stream" in https://github.com/whatwg/streams/pull/871, so those won't be needed once Bikeshed has picked up the change.

encoding.bs
ricea: I've added "then" to all instances of ", set" and ", return". I haven't (intentionally) changed any that weren't in my patch.

I still have some instances of the pattern:

> If <var>token</var> is <a>end-of-stream</a>, run these steps:

Are these okay, or should I add "then" here as well?

encoding.bs
ricea: That's really useful, thanks.

encoding.bs
annevk: Shall we make Fetch stop doing that then? It doesn't really seem like Fetch's way of managing it handles anything.

encoding.bs
annevk: I prefer then there too. At some point I should do an editorial change for the rest of this document.

encoding.bs
ricea: I get it now, thank you.

ricea:
@hsivonen wrote:
> Please add a very conspicuous note explaining that unlike in cases where the JS string to DOMString boundary passes through data that doesn't conform to Unicode as a performance optimization / legacy artifact of failure to enforce Unicode conformance, in this case letting code unit sequences that do not conform to Unicode travel across the boundary is an explicit feature with the intent of allowing the sequence of code units to become conforming once concatenated on the other side of the JS string / DOMString boundary.

I'm concerned that this may contradict the [advice from the WebIDL standard](https://heycam.github.io/webidl/#idl-USVString), namely:

> Specifications should only use USVString for APIs that perform text processing and need a string of Unicode scalar values to operate on. Most APIs that use strings should instead be using DOMString, which does not make any interpretations of the code units in the string. When in doubt, use DOMString.

I'd prefer to have some clarity about whether there is a contradiction here and what the resolution should be before adding an editorial note.

encoding.bs
ricea: @yutakahirano Who is using the ReadableStream definition that is exported from Fetch?

ricea:
@annevk I had completely forgotten what I agreed to in #72. Thanks for catching that. There's quite a few extra tests needed so it won't be ready until next week.

encoding.bs
yutakahirano: I find none, but serviceworker uses related ones ("construct a ReadableStream object", "empty ReadableStream"). See also: https://github.com/heycam/webidl/issues/445

ricea:
I've started a side-discussion about how to best implement mode-locking at https://github.com/whatwg/streams/issues/873. The question is basically whether to add some complexity to the Streams Standard to reduce some complexity in the Encoding Standard.

ricea:
FYI: I've added tests that objects are created in the correct realms: https://github.com/GoogleChromeLabs/text-encode-transform-prollyfill/pull/11

ricea:
I've attempted to implement the "Constructor version" of the API discussed https://github.com/whatwg/encoding/issues/72#issuecomment-372601600. It doesn't seem too bad.

I have some tests in https://github.com/GoogleChromeLabs/text-encode-transform-prollyfill/blob/f4de56ca18b98e55a79e7f2c9cb7658097b2aaf4/tests/non-interference.html. From the perspective of a consumer of the API, the behaviour looks pleasantly unsurprising to me.

This is not a final decision to go with the "Constructor version" of the API. I could switch to the "Static method version" depending on how people think this looks.

encoding.bs
```diff
@@ -1314,6 +1528,120 @@ must run these steps:
...
+
+<ol>
+ <li><p>Let <var>input</var> be the result of <a lt="converted to an IDL value">converting</a>
+ <var>chunk</var> to a {{DOMString}}. If this throws an exception, then return a promise rejected
+ with that exception.
+
+ <p class=note>{{DOMString}} is used here so that a surrogate pair that is split between chunks can
+ be correctly reassembled into the appropriate code point in the output. The behaviour is otherwise
+ identical to {{USVString}}. In particular, replacement characters will be used where necessary to
+ make the output valid UTF-8.
```
annevk: I'd prefer if we stated that lone surrogates will be replaced with U+FFFD.

ricea: Thanks. How is the updated version? I'm not sure whether "valid UTF-8" or "well-formed UTF-8" is better.

annevk: I guess I'd just say "is UTF-8", but for a note it seems fine. Although having said that, it's a little weird that it both talks about code points in the output and says the output is UTF-8 (which is bytes).

annevk: How about:

> {{DOMString}} is used here so that a surrogate pair that is split between chunks can be reassembled into the appropriate scalar value. The behaviour is otherwise identical to {{USVString}}. In particular, lone surrogates will be replaced with U+FFFD.

ricea:
Closing in favour of #149.

PR #130, opened 2017-12-19
Meta: update .travis.yml

foolip:
To adapt to https://github.com/whatwg/whatwg.org/pull/119 and https://github.com/whatwg/whatwg.org/pull/137.


foolip:
OK, experimenting is done, this is now just a plain PR that can be merged when convenient.

PR #136, opened 2018-04-25
Be clearer on which user agents the API applies to

### LINKED ISSUE #87: Non-browser versus JavaScript
Currently we state "Non-browser user agents are not required to support this API."

I wonder if we should phrase that differently somehow to be inclusive of other JavaScript deployments, such as Node.js. @jasnell thoughts? I'm guessing Node.js has its own encoding libraries, but maybe this is another place where we can pursue convergence.

annevk:
Fixes #87.

***
<a href="https://whatpr.org/encoding/136.html" title="Last updated on Apr 25, 2018, 12:05 PM GMT (b4e5aad)">Preview</a> | <a href="https://whatpr.org/encoding/136/9c80222...b4e5aad.html" title="Last updated on Apr 25, 2018, 12:05 PM GMT (b4e5aad)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -966,10 +966,9 @@ return the result of <a lt=encode for=/>encoding</a>
 
 <h2 id=api>API</h2>
 
-<p>This section uses terminology from Web IDL. Non-browser user agents are not required to support
-this API.
-
-[[!WEBIDL]]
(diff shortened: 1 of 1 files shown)


annevk:
@jasnell what do you think?

jasnell:
Lgtm

PR #140, opened 2018-05-12
Include the character whose pointer is zero in BMP coverage visualizations

hsivonen:
Currently, the character whose pointer is zero shows up as unassigned in the BMP coverage visualizations. Let's fix that.


hsivonen:
> No index maps to U+0000 I suppose,

Initially I didn't change it for this reason, but I added another changeset for completeness.

PR #143, opened 2018-05-14
Add Streams Explainer

ricea:
Add a new file, streams-explainer.md, which explains the integration of
Streams into the Encoding Standard for a general audience.

### DIFF
--- a/stream-explainer.md
+++ b/stream-explainer.md
@@ -1,0 +1,206 @@
+# Encoding Streams Explained
+14 May 2018
+
+## What’s all this then?
+
+**TextDecoderStream** is used to convert a stream of text in a binary encoding to a
+[stream](https://streams.spec.whatwg.org/) of strings.
+
+**TextEncoderStream** is used to convert a stream of strings into bytes in the UTF-8 encoding.
+
+Both APIs satisfy the concept of a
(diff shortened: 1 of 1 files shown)


annevk:
This seems great, but perhaps better suited as a blog post? @domenic what do you think about putting this on blog.whatwg.org?

domenic:
I think there are two audiences here. One is folks who appreciate explainers, like spec reviewers and blink-dev. The other is web developers. The content won't be too different in each case, but I'd imagine for the first audience we want to couch this more as a proposal, whereas for the latter audience we want to publish an announcement of a new feature. And the latter audience might want to be linked to some of the gory details, instead of consuming it all inside their feed reader.

So I think we should keep the explainer as-is, either in this repo or in @ricea's fork, for folks who appreciate explainers. And after we've successfully landed the spec change, we can work on turning it into a blog post.

WDYT?

annevk:
Okay, I guess I don't quite see explainers being part of Living Standards or their repositories. If a feature is sufficiently unclear it should just get an introduction section or some such.

ricea:
@annevk Thanks, I will withdraw this pull request.

Hopefully the feature won't need much explanation in the standard itself. I expect a couple of examples will be enough.

PR #145, opened 2018-05-29
Meta: Review Draft infrastructure

annevk:
See https://github.com/whatwg/whatwg.org/issues/197 and https://github.com/whatwg/meta/issues/92 for details.

This also updates the web-platform-tests and IRC URLs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
+/UnicodeData.txt
 /encoding.spec.whatwg.org/
-deploy_key
-deploy_key.pub
-encoding.html
-UnicodeData.txt
-deploy.sh
-review.sh
+/deploy_key
+/deploy_key.pub
+/deploy.sh
+/encoding.html
+/review.sh
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ deploy_key
 deploy_key.pub
 encoding.html
 UnicodeData.txt
+deploy.sh
+review.sh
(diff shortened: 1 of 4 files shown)


domenic:
Agreed, then you can successfully deploy using the Makefile

annevk:
Hmm, I cannot get
```make
deploy: encoding.bs
	export CHECKER_FILTER="Text run is not in Unicode Normalization Form C\.|.*appears to be written in.*|.*Charmod C073.*"
	export EXTRA_FILES="*.txt *.json *.css"
	export POST_BUILD_STEP='python visualize.py "$DIR/"'
	curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh && bash ./deploy.sh
```
to work. Also when I omit `export `.

It also does not work for Console for me locally. Perhaps that's just a bug in `make` on macOS?

domenic:
Hmm, I'm not a Make expert, although https://stackoverflow.com/q/23843106 has a number of interesting things... I was just going to do this though (note the line-ending `\`s):

```makefile
deploy: encoding.bs
  curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh
	CHECKER_FILTER="Text run is not in Unicode Normalization Form C\.|.*appears to be written in.*|.*Charmod C073.*" \
	EXTRA_FILES="*.txt *.json *.css" \
	POST_BUILD_STEP='python visualize.py "$DIR/"' \
	bash ./deploy.sh
```

annevk:
That works better, thanks. Now the problem is that $D is removed.

annevk:
Ah, $$ per https://stackoverflow.com/questions/7654386/how-do-i-properly-escape-data-for-a-makefile.

PR #149, opened 2018-07-18
Add TextEncoderStream and TextDecoderStream transform streams

### LINKED ISSUE #72: Add Streams support

ricea:
Integrate with the streams standard by adding TextEncoderStream and
TextDecoderStream transform streams to the standard. These enable
binary<>text conversions on a ReadableStream using the `pipeThrough()`
method (see https://streams.spec.whatwg.org/#rs-pipe-through).

A TextEncoderStream object can be used to transform a stream of strings
to a stream of bytes in UTF-8 encoding. A TextDecoderStream object can
be used to transform a stream of bytes in the encoding passed to the
constructor to strings.

There is a prollyfill and tests for the new functionality at
https://github.com/GoogleChromeLabs/text-encode-transform-prollyfill.

Closes #72.

***
<a href="https://whatpr.org/encoding/149.html" title="Last updated on Aug 14, 2018, 1:45 PM GMT (b9dfb01)">Preview</a> | <a href="https://whatpr.org/encoding/149/19af047...b9dfb01.html" title="Last updated on Aug 14, 2018, 1:45 PM GMT (b9dfb01)">Diff</a>


encoding.bs
```diff
@@ -15,6 +15,19 @@ Translate IDs: dictdef-textdecoderoptions textdecoderoptions,dictdef-textdecodeo
 spec:infra; type:dfn;
     text:code point
     text:ascii case-insensitive
+spec:streams;
+    type:interface; text:ReadableStream
+    type:dfn; text:chunk
```
domenic: We should export these in streams so that this is not necessary, for the dfns at least. For the interface, that's a separate issue: https://github.com/whatwg/fetch/issues/780 so we can leave it here for now.

encoding.bs
```diff
@@ -1038,6 +1051,33 @@ function decodeArrayOfStrings(buffer, encoding) {
...
+};
+</pre>
+
+<p>The {{TextDecoderAttributes}} mixin defines common attributes that are shared between
+{{TextDecoder}} and {{TextDecoderStream}}. These objects have an associated
+<dfn for=TextDecoderAttributes>encoding</dfn>, <dfn for=TextDecoderAttributes>ignore BOM flag</dfn>,
+and <dfn for=TextDecoderAttributes>error mode</dfn>.
+
+<p>The <dfn attribute for=TextDecoderAttributes><code>encoding</code></dfn> attribute's getter must
+return <a for=TextDecoderAttributes>encoding</a>'s <a for=encoding>name</a> in <a>ASCII
```
domenic: While moving things around maybe use the modern phrasing "this object's encoding's" or "this TextDecoderAttribute object's encoding's". Similarly for the rest.

encoding.bs
```diff
@@ -1038,6 +1051,33 @@ function decodeArrayOfStrings(buffer, encoding) {
 </div>
 
 
+<h3 id=interface-mixin-textdecoderattributes>Interface Mixin {{TextDecoderAttributes}}</h3>
```
domenic: Nit: lowercase "mixin" in headings, here and below. (Headings are sentence case.)

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+
+<pre class=idl>
+interface mixin GenericTransformStream {
+  readonly attribute ReadableStream readable;
+  readonly attribute WritableStream writable;
+};
+</pre>
+
+<p>An object that includes GenericTransformStream always has an associated
+<dfn for=GenericTransformStream>transform</dfn>.
```
domenic: Probably worth stating its type.

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+a {{TransformStream}}, though it has the same interface and it delegates to one.
+
+<pre class=idl>
+interface mixin GenericTransformStream {
+  readonly attribute ReadableStream readable;
+  readonly attribute WritableStream writable;
+};
+</pre>
+
+<p>An object that includes GenericTransformStream always has an associated
```
domenic: `{{GenericTransformStream}}`

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+TextDecoderStream includes GenericTransformStream;
+</pre>
+
+<p>A {{TextDecoderStream}} object has an associated <dfn for=TextDecoderStream>encoding</dfn>,
+<dfn for=TextDecoderStream>decoder</dfn>, <dfn for=TextDecoderStream>stream</dfn>,
+<dfn for=TextDecoderStream>ignore BOM flag</dfn> (initially unset),
+<dfn for=TextDecoderStream>BOM seen flag</dfn> (initially unset),
+<dfn for=TextDecoderStream>error mode</dfn> (initially "<code>replacement</code>"),
+<dfn for=TextDecoderStream>do not flush flag</dfn> (initially unset), and
+<dfn for=TextDecoderStream>transform</dfn>.
```
domenic: For those things provided by the mixins, you should use the mixin `dfn`s, and for=TextDecoderAttributes on references to them. No need to have separate dfns/internal slots for the TextDecoder and TextDecoderStream themselves.

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+
+<p>A {{TextDecoderStream}} object also has an associated <dfn id=concept-tds-serialize
+for=TextDecoderStream>serialize stream</dfn> algorithm, that is identical to the <a
+for=TextDecoder>serialize stream</a> algorithm for the equivalent {{TextDecoder}}.
```
domenic: Can this move to TextDecoderAttributes since it's shared? Not sure myself.

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+ <dt><code><var>decoder</var> . <a attribute for=GenericTransformStream>writable</a></code>
+ <dd>
+  <p>Returns a <a>writable stream</a> which accepts {{BufferSource}} chunks and runs them through
+  <a for=TextDecoderStream>encoding</a>'s <a for=/>decoder</a> before making them available to
+  {{GenericTransformStream/readable}}.
+
+  <p>Typically this will be used via the {{ReadableStream/pipeThrough()}} method on a
+  {{ReadableStream}} source.
+
+  <pre class=example id=example-textdecoderstream-writable><code class=lang-javascript>
```
domenic: FWIW in Bikeshed you can use any indentation/whitespacing you want.

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+{{TextDecoderStream}} <var>dec</var> and a <var>chunk</var>, runs these steps:
+
+<ol>
+ <li><p>If <a>Type</a>(<var>chunk</var>) is not Object, or <var>chunk</var> does not have an
+ \[[ArrayBufferData]] internal slot, or <a abstract-op>IsDetachedBuffer</a>(<var>chunk</var>) is
+ true, or <a abstract-op>IsSharedArrayBuffer</a>(<var>chunk</var>) is true, then return a new
+ promise rejected with a {{TypeError}} exception.
+
+ <li><p><a>Push</a> a <a lt="get a copy of the buffer source">copy of</a> <var>chunk</var> to
+ <var>dec</var>'s <a for=TextDecoderStream>stream</a>.
```
domenic: Hmm one day we should do a transfer variant, of this and all other web platform APIs...

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+</pre>
+
+<p>A {{TextEncoderStream}} object has an associated <dfn for=TextEncoderStream>encoder</dfn>,
+<dfn for=TextEncoderStream>transform</dfn> and <dfn for=TextEncoderStream>pending high
+surrogate</dfn> (initially null).
+
+<p class="note no-backref">A {{TextEncoderStream}} object offers no <var>label</var> argument as it
+only supports <a>UTF-8</a>.
+
+<hr>
```
domenic: Not so sure about this `<hr>`

encoding.bs
```diff
@@ -1305,6 +1350,404 @@ must run these steps:
...
+ surrogate</a> to <var>token</var> and return <a>continue</a>.
+
+ <li><p>If <var>token</var> is in the range U+DC00 to U+DFFF, inclusive, then return U+FFFD.
+
+ <li><p>Return <var>token</var>.
+</ol>
+
+<p class=note>This is equivalent to the "<a>convert</a> a <a>JavaScript string</a> into a <a>scalar
+value string</a>" algorithm from the Infra Standard, but allows for surrogate pairs that are split
+between strings.
```
domenic: Add `[[INFRA]]` for a nice stylistic flourish

encoding.bs
ricea: I've added it to the accessors, but I will probably need to add it some of the algorithms as well.

domenic:
Did a re-review and it all looks great, pending further explorations into sharing things more (especially with regard to the encoder classes).

encoding.bs
```diff
@@ -1305,6 +1348,398 @@ must run these steps:
...
+  <ol>
+   <li><p>Let <var>token</var> be the result of <a>reading</a> from <var>input</var>.
+
+   <li>
+    <p>If <var>token</var> is <a>end-of-stream</a>, then run these steps:
+
+    <ol>
+     <li><p>Convert <var>output</var> into a byte sequence.
+
+     <li><p>Let <var>chunk</var> be a {{Uint8Array}} object wrapping an {{ArrayBuffer}} containing
```
MattiasBuelens: Once again, skip this step and the following enqueue when `output` is an empty byte sequence. For example, this could happen if `chunk` is empty, but also if `chunk` consists of one single high surrogate.

ricea:
@MattiasBuelens Thank you for the review.

> We should avoid enqueuing empty strings (in TextDecoderStream) or empty Uint8Arrays (in TextEncoderStream).

I'm not completely sure about this. I originally preserved empty chunks on the assumption that users would expect that if they put an empty chunk in they would get an empty chunk out.

I implemented this at https://github.com/GoogleChromeLabs/text-encode-transform-prollyfill/compare/never-empty?expand=1. Please take a look at the test changes and see what you think of the changed semantics.

MattiasBuelens:
@ricea I think I prefer the updated semantics.

I can't really think of a use case where you'd be interested in empty chunks/strings. I feel like it'd just make user code unnecessarily more complicated, e.g.:
```js
stream
    .pipeThrough(new TextDecoderStream())
    .pipeThrough(new TransformStream({
        transform(chunk, controller) {
            if (chunk.byteLength === 0) {
                return; // what else would you do here?
            }
            // do stuff
        }
    }));
```

I'm not aware of any spec that enqueues empty chunks to a stream. [Fetch doesn't do it](https://fetch.spec.whatwg.org/#ref-for-concept-enqueue-readablestream%E2%91%A0):
> 12.1.1.1. **If one or more bytes have been transmitted** from response’s message body, then:
> 12.1.1.1.1. Let bytes be the transmitted bytes.
> 12.1.1.1.2. [...]
> 12.1.1.1.6. **Enqueue** a `Uint8Array` object wrapping an `ArrayBuffer` containing bytes to stream. [...]

MattiasBuelens:
Regardless of whether we decide to enqueue an empty chunk/string inside _decode/encode and enqueue a chunk_, we **definitely** should not enqueue an empty chunk inside _flush and enqueue_. 😉

encoding.bs
```diff
@@ -1305,6 +1348,398 @@ must run these steps:
...
+
+ <dt><code><var>decoder</var> . <a attribute for=GenericTransformStream>readable</a></code>
+ <dd>
+  <p>Returns a <a>readable stream</a> whose <a>chunks</a> are strings resulting from running
+  <a for=TextDecoderAttributes>encoding</a>'s <a for=/>decoder</a> on the chunks written to
+  {{GenericTransformStream/writable}}.
+
+ <dt><code><var>decoder</var> . <a attribute for=GenericTransformStream>writable</a></code>
+ <dd>
+  <p>Returns a <a>writable stream</a> which accepts {{BufferSource}} chunks and runs them through
```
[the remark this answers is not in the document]
ricea: Thanks for catching this. My attempt to fix it (6a65ad5) is ugly, but hopefully between us we can find a way to say it nicely.

domenic:
I think for ArrayBufferView we can lean on Web IDL more:

> Let _bufferSource_ be the result of [converting](https://heycam.github.io/webidl/#dfn-convert-ecmascript-to-idl-value) _chunk_ to a `BufferSource`.

Then "a copy of" works well.

encoding.bs
```diff
@@ -1305,6 +1349,399 @@ must run these steps:
...
+
+<p>The <dfn id=concept-tds-decode-and-enqueue>decode and enqueue a chunk</dfn> algorithm, given a
+{{TextDecoderStream}} <var>dec</var> and a <var>chunk</var>, runs these steps:
+
+<ol>
+ <li><p>Let <var>bufferSource</var> be the result of
+ <a lt="converted to an IDL value">converting</a> <var>chunk</var> to a {{BufferSource}}. If this
+ throws an exception, then return a promise rejected with that exception.
+
+ <li><p><a>Push</a> a <a lt="get a copy of the buffer source">copy of</a> <var>bufferSource</var> to
```
MattiasBuelens: In 6a65ad5da12cd43b24ee9459ea793d285f424912, we explicitly checked for `IsDetachedBuffer(arrayBuffer)` and returned a rejected promise if the buffer was detached. However, none of the three algorithms that make up [_convert to a BufferSource_](https://heycam.github.io/webidl/#es-buffer-source-types) check for detached buffers.

When [_get a copy of the buffer source_](https://heycam.github.io/webidl/#ref-for-dfn-get-buffer-source-copy%E2%91%A0) encounters a detached buffer, it throws a `TypeError` in step 7. This is not what we want: we want to return a rejected promise instead!

I think the easiest fix would be to add the following line to this step, similar to the previous step:
> If this throws an exception, then return a promise rejected with that exception.

ricea: Great catch, thanks.

I did what you proposed. This has the added benefit that we are always in sync with the checks and behaviour provided by WebIDL.

encoding.bs
```diff
@@ -1485,7 +1485,8 @@ byteReadable
  throws an exception, then return a promise rejected with that exception.
 
  <li><p><a>Push</a> a <a lt="get a copy of the buffer source">copy of</a> <var>bufferSource</var> to
- <var>dec</var>'s <a for=TextDecoderStream>stream</a>.
+ <var>dec</var>'s <a for=TextDecoderStream>stream</a>. If this throws an exception, then return a
+ promise rejected with that exception
```
MattiasBuelens: Tiny nit: missing a period at the end of this sentence. 😛

annevk:
I trust that you got the various technical aspects correctly here, especially after the careful review from others. I would like to ask you though that for the various `<dfn>` elements which you have changed the `for` attribute of (as needed) to give them an `id` attribute to preserve their existing ID. Not breaking those links seems worthwhile.

I also pushed a commit that fixes various other things I noticed while looking through the text.

annevk:
I suggest we also add @MattiasBuelens to the Acknowledgments section for their help in reviewing.

ricea:
> I would like to ask you though that for the various <dfn> elements which you have changed the for attribute of (as needed) to give them an id attribute to preserve their existing ID.

Please take a look at af33687 and let me know whether I have done this correctly.

hsivonen:
Is there a reason why `TextDecoderStream` doesn't provide a mode with the semantics of the "decode" algorithm for BOM handling (i.e. use the _encoding_ if there's no BOM but honor the BOM if there is one)? These semantics are necessary to consume `text/plain` and these semantics are hard for developers to get right in the streaming context.

Also, does one input chunk always result in one output chunk that corresponds to potential pending partial code unit sequence and the input chunk except for potential partial code unit sequence at the end of the input chunk? (If not, I'm a bit worried about Web content developing a dependency on browser-specific chunk boundaries where the boundaries are not supposed to mean anything.)

ricea:
> Is there a reason why TextDecoderStream doesn't provide a mode with the semantics of the "decode" algorithm for BOM handling

The short answer is "because TextDecoder doesn't". I don't know the historical reason why TextDecoder doesn't.

I agree that BOM sniffing is necessary to parse legacy content, and that it can be hard to get right. However, I'd like to discourage people from relying on this behaviour. I'd like to move away from any kind of heuristic behaviour and towards a world where everyone uses UTF-8, and that's what the default behaviour encourages.

If we find compelling use cases for easily reading legacy content we may need to support it in future, but my personal preference would be to leave it out until it is proven necessary.

> Also, does one input chunk always result in one output chunk that corresponds to potential pending partial code unit sequence and the input chunk except for potential partial code unit sequence at the end of the input chunk?

Input chunk to output chunk correspondance is normally 1:1, except that we don't output empty chunks, and an extra chunk may be output at the end of the stream if we discover the input was incomplete.

>  (If not, I'm a bit worried about Web content developing a dependency on browser-specific chunk boundaries where the boundaries are not supposed to mean anything.)

The semantics are strictly greedy. Given the same chunks as input, every browser is required to have exactly the same chunk boundaries in the output.

ricea:
@annevk Sorry that stray call to `decode()` crept in. I just re-checked against the IDL we are testing in the idlharness.html test and it matches. Unfortunately I can't just copy-and-paste because idlharness.js doesn't support mixins.

annevk:
@ricea it does support mixins I think? Lots of IDL in https://github.com/web-platform-tests/wpt/tree/master/interfaces uses them.

ricea:
@annevk Oh, I should have tried it!

I've switched my work-in-progress copy to using the IDL from the standard verbatim and it does work.

hsivonen:
> If we find compelling use cases for easily reading legacy content we may need to support it in future, but my personal preference would be to leave it out until it is proven necessary.
...
> The semantics are strictly greedy. Given the same chunks as input, every browser is required to have exactly the same chunk boundaries in the output.

OK. Thanks.

ricea:
@annevk I have been refining the tests. Would you like me to land them before landing this?

annevk:
Great, just need a pointer to them, ideally they aren't merged until this is fully agreed on.

The other thing that's remaining here is implementer bugs: https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests.

ricea:
Tests are at https://github.com/web-platform-tests/wpt/pull/12430.

The Chrome bug at least exists: https://bugs.chromium.org/p/chromium/issues/detail?id=845427 :smiley:

annevk:
I pushed some further nits. I think I can say that Mozilla is on board with this PR (per Henri, Till, and I), which brings us to two implementers.

@othermaciej @travisleithead any final feedback / concerns?

@ricea would you file implementation bugs against the other implementation as per my link above?

travisleithead:
Looked it over and don't have any concerns. Great work! I searched for a bug in our external issue tracker, and didn't see one, but we do have Encoding API (the whole spec) noted as In Development, and you can perhaps track this request via: https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/6558040-support-the-encoding-api if it helps :).

ricea:
Sorry for the delay, I was on holiday.

Firefox issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1486949
Safari issue: https://bugs.webkit.org/show_bug.cgi?id=189066

I'm going to assume that https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/6558040-support-the-encoding-api covering the whole Encoding API is sufficient for Edge, at least until they have some of it implemented.

PR #151, opened 2018-08-23
Add an informative note about the relationship of encodings and Unicode encoding schemes

hsivonen:
Seems useful to explain the difference between concepts here and in Unicode.

***
<a href="https://whatpr.org/encoding/151.html" title="Last updated on Aug 23, 2018, 1:38 PM GMT (433e152)">Preview</a> | <a href="https://whatpr.org/encoding/151/19af047...433e152.html" title="Last updated on Aug 23, 2018, 1:38 PM GMT (433e152)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -135,6 +135,16 @@ a <a>byte</a> sequence (and vice versa). Each <a for=/>encoding</a> has a
 <dfn id=name export for=encoding>name</dfn>, and one or more
 <dfn id=label export for=encoding lt=label>labels</dfn>.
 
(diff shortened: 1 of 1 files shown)


annevk:
Thanks, I pushed some nits. In particular, we already had a note about this for UTF-16, but not UTF-8, so I added one there too. The duplication seems reasonable to me as this is probably a source of some confusion. And since this is now the first mention of the byte order mark I updated the other mention to not say BOM.

hsivonen:
> Thanks, I pushed some nits

Looks good to me. Thanks.

PR #155, opened 2018-08-30
ISO-2022-JP encoder: document an oddity

### LINKED ISSUE #115: Concatenating two ISO-2022-JP outputs from a conforming encoder doesn't result in conforming input
Encodings other than ISO-2022-JP have the property that if you concatenate two outputs from a conforming encoder and decode them together, you get the same result as when decoding them separately and then concatenating.

ISO-2022-JP lacks this property, because despite the encoder making an effort into this direction by emitting a transition to the ASCII state at the end, if the next segment being concatenated starts with a transition to a non-ASCII state, the concatenation results in zero ASCII bytes between two escapes.

Is there a strong reason for treating a transition to the ASCII state immediately followed by another escape as non-conforming? Or put the other way, what purpose does the transition to the ASCII state at the end of encode serve if not achieving the above-mentioned concatenation property that other encodings have?

[This is relevant to RFC 2047 header decoding](https://bugzilla.mozilla.org/show_bug.cgi?id=1374149).

annevk:
At this point it does not seem worth it to require further implementation changes and risk compatibility issues, so instead document the quirk.

Closes #115.

***
<a href="https://whatpr.org/encoding/155.html" title="Last updated on Sep 2, 2018, 9:31 AM GMT (9efea42)">Preview</a> | <a href="https://whatpr.org/encoding/155/b579018...9efea42.html" title="Last updated on Sep 2, 2018, 9:31 AM GMT (9efea42)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -2650,6 +2650,15 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
 
 <h4 id=iso-2022-jp-encoder dfn export>ISO-2022-JP encoder</h4>
 
+<div class="note no-backref">
+ <p>The <a>ISO-2022-JP encoder</a> is the only <a for=/>encoder</a> for which the concetenation of
+ multiple outputs can result in an <a>error</a> when run through the <a>ISO-2022-JP decoder</a>.
+
+ <p class=example id=example-iso-2022-jp-encoder-oddity>Encoding U+00A5 gives 0x1B 0x28 0x4A 0x5C
+ 0x1B 0x28 0x42. Doing that twice, concatenating the results, and then decoding yields U+00A5 U+FFFD
+ U+00A5.
+</div>
+
 <p><a>ISO-2022-JP</a>'s <a for=/>encoder</a> has an associated
 <dfn>ISO-2022-JP encoder state</dfn> which is <dfn lt="ISO-2022-JP encoder ASCII">ASCII</dfn>,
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -2650,6 +2650,15 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
 
 <h4 id=iso-2022-jp-encoder dfn export>ISO-2022-JP encoder</h4>
 
+<div class="note no-backref">
+ <p>The <a>ISO-2022-JP encoder</a> is the only <a for=/>encoder</a> for which the concetenation of
```
domenic: concetenation -> concatenation

encoding.bs
```diff
@@ -2650,6 +2650,15 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
 
 <h4 id=iso-2022-jp-encoder dfn export>ISO-2022-JP encoder</h4>
 
+<div class="note no-backref">
+ <p>The <a>ISO-2022-JP encoder</a> is the only <a for=/>encoder</a> for which the concatenation of
+ multiple outputs can result in an <a>error</a> when run through the <a>ISO-2022-JP decoder</a>.
```
hsivonen: I'd say "corresponding decoder" instead of "ISO-2022-JP decoder" to be pedantic about this being self-decode and not about decoding the output of other encoders with the ISO-2022-JP decoder. Other than that, looks good to me.

PR #166, opened 2018-12-13
Define encodeInto() API

### LINKED ISSUE #69: TextEncoder#encode - write to existing Uint8Array

annevk:
This enables converting strings into byte sequences of pre-allocated buffers.

Also cleanup TextEncoder a bit.

Tests: ...

Fixes #69.

***
<a href="https://whatpr.org/encoding/166.html" title="Last updated on Jan 8, 2019, 2:04 PM UTC (7bb77bb)">Preview</a> | <a href="https://whatpr.org/encoding/166/c2879dd...7bb77bb.html" title="Last updated on Jan 8, 2019, 2:04 PM UTC (7bb77bb)">Diff</a>


annevk:
The reason this does not contain the StringView dictionary is because it can be added later if there's indeed a need for it. This also does not require any particular destination length as either way out-of-band needs to be dealt with and it seems better to be flexible for simple string use cases that would otherwise have to write completely different code.

hsivonen:
Looks correct to me, but it took me a moment to realize that the algorithm always loops over all tokens and the loop just does not necessarily write all the output. While an early return might be just an optimization, it might make the algorithm more comprehensible.

annevk:
Thanks for pointing that out, I think it's actually wrong. Say we had 3 bytes left and the next input required 4 bytes and the input after that 1 byte. This would skip the 4 bytes input and then write the 1 byte one.

encoding.bs
```diff
@@ -1319,18 +1323,14 @@ requires buffering of scalar values.
 
  <dt><code><var>encoder</var> . <a method for=TextEncoder lt=encode()>encode([<var>input</var> = ""])</a></code>
  <dd><p>Returns the result of running <a>UTF-8</a>'s <a for=/>encoder</a>.
+
+ <dt><code><var>encoder</var> . <a method=for=TextEncoder lt="encodeInto(source, destination)">encodeInto(<var>source</var>, <var>destination</var>)</a></code>
+ <dd><p>Runs the <a>UTF-8 encoder</a> on <var>source</var>, stores the result on
```
ricea: "storing the result into destination" would sound more natural to me.

annevk:
Do you know what the wording for those places should be? IDL doesn't really define any language around IDL ArrayBuffer objects... I guess I can file an issue upstream at least.

The other thing I realized is that I should add at least 1 example and document more clearly what read and written represent.

inexorabletash:
I think this is implied by your last comment, but just to confirm: as written, _destination_'s view bounds (offset, length) onto its ArrayBuffer is seemingly ignored by the current algorithm. i.e. in this case:
```js
const ab = new ArrayBuffer(5);
const ua = new Uint8Array(ab, 1, 2);
new TextEncoder().encodeInto("ABCDE", ua);
```
... I would expect `ab` to end up as [0, 0x65, 0x66, 0, 0]. 

(I don't see test cases for this either)

annevk:
Yeah, that clearly needs to be worded better. (I added tests.)

annevk:
I think this is good now apart from the example which I'd like feedback on over at https://github.com/whatwg/encoding/issues/69#issuecomment-447275930. (And there's a Bikeshed issue I'll resolve with Tab.)

annevk:
I haven't explained why there's no `decodeInto()`. I suppose "bytes in encoding X" -> "bytes in UTF-8" could make sense, but let's only add that if there's a request.

Anyone want to do a final review pass? (Tests need one too.)

encoding.bs
```diff
@@ -1347,20 +1350,108 @@ must run these steps:
...
-    byte sequence, and then return a {{Uint8Array}} object wrapping an
-    {{ArrayBuffer}} containing <var>output</var>.
-    <!-- XXX https://www.w3.org/Bugs/Public/show_bug.cgi?id=26966 -->
+    <p>Otherwise:
 
-    <p class=note><a>UTF-8</a> cannot return <a>error</a>.
+    <ol>
+     <li>
+      <p>If <var>destinationBytes</var>'s <a for="byte sequence">length</a> &minus;
+      <var>written</var> is greater than the number of bytes in <var>result</var>, then:
```
ricea: I think this needs to be "greater than or equal to".

annevk: Yeah, indeed, thanks! (E.g., if we have space for 12 bytes, already written 6 bytes, then writing 6 more bytes should be fine whereas the current wording prohibits that.)

annevk:
I'll land this and the tests tomorrow, unless there are further comments.

annevk:
cc @whatwg/documentation

annevk:
Implementation bugs:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1514664
* https://bugs.webkit.org/show_bug.cgi?id=193274
* https://bugs.chromium.org/p/chromium/issues/detail?id=920107

Issue #154, opened 2018-08-30
TextDecoder's serialize stream algorithm seems unnecessarily convoluted.


Codifier:
It appears to me that some algorithms are unnecessarily convoluted. One in particular is the [TextDecoder's serialize stream algorithm](https://encoding.spec.whatwg.org/commit-snapshots/6f9a41f3d9dbc7ba6d88f65f7ef1c139fb08d4be/#concept-td-serialize):

> 1. Let output be the empty string.
> 2. While true:
>    1. Let token be the result of reading from stream.
>    2. If encoding is UTF-8, UTF-16BE, or UTF-16LE, and ignore BOM flag and BOM seen flag are unset, then:
>       1. If token is U+FEFF, then set BOM seen flag.
>       2. Otherwise, if token is not end-of-stream, then set BOM seen flag and append token to output.
>       3. Otherwise, return output.
>    3. Otherwise, if token is not end-of-stream, then append token to output.
>    4. Otherwise, return output.

Unless I'm mistaken, I believe it could be simplified to:

> 1. Let output be the empty string.
> 2. While true:
>    1. Let token be the result of reading from stream.
>    2. If token is end-of-stream, return output.
>    3. If encoding is UTF-8, UTF-16BE, or UTF-16LE, and ignore BOM flag and BOM seen flag are unset, then:
>       1. Set BOM seen flag.
>       2. If token is U+FEFF, then continue.
>    4. Append token to output.

...because the "BOM seen flag" will only ever be set if the token is *not* "end-of-stream".

ricea:
This simplification appears to be correct to me. I'd like some other people to double-check it.

It's also quite close to what Blink actually implements, but since Blink's implementation is layered differently, there isn't a direct 1:1 mapping.

annevk:
I looked at blame and the original algorithm is 5 years old. I'm pretty sure that was before we standardized on "continue" on Infra, so perhaps that explains the way it is. In any event, rewriting it seems acceptable.

@Codifier how would you like to be acknowledged?

Codifier:
@annevk You can mention me as CodifierNL in any acknowledgement you deem appropriate. Thank you.

is2ei:
@annevk 
Anyone works on this issue? If not, I'd like to work on this.

annevk:
@is2ei please do!

PR #175, opened 2019-04-11
Editorial: use IDL's this and adopt more Infra

annevk:
And cleanup various other editorial patterns.

***
<a href="https://whatpr.org/encoding/175.html" title="Last updated on May 13, 2020, 4:30 AM UTC (d1af7ea)">Preview</a> | <a href="https://whatpr.org/encoding/175/b991dea...d1af7ea.html" title="Last updated on May 13, 2020, 4:30 AM UTC (d1af7ea)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -1376,7 +1376,7 @@ requires buffering of scalar values.
 
  <dt><code><var>encoder</var> . <a method=for=TextEncoder lt="encodeInto(source, destination)">encodeInto(<var>source</var>, <var>destination</var>)</a></code>
  <dd><p>Runs the <a>UTF-8 encoder</a> on <var>source</var>, stores the result of that operation into
- <var>destination</var>, and returns the progress made as a dictionary whereby
+ <var>destination</var>, and returns the progress made as an object whereby
  {{TextEncoderEncodeIntoResult/read}} is the number of converted <a>code units</a> of
  <var>source</var> and {{TextEncoderEncodeIntoResult/written}} is the number of bytes modified in
  <var>destination</var>.
@@ -1474,9 +1474,8 @@ method steps are:
     </ol>
   </ol>
 
- <li><p>Return a new {{TextEncoderEncodeIntoResult}} dictionary whose
- {{TextEncoderEncodeIntoResult/read}} member is <var>read</var> and
- {{TextEncoderEncodeIntoResult/written}} member is <var>written</var>.
+ <li><p>Return «[ "{{TextEncoderEncodeIntoResult/read}}" → <var>read</var>,
+ "{{TextEncoderEncodeIntoResult/written}}" → <var>written</var> ]».
 </ol>
 
 <div class=example id=example-textencoder-encodeinto>
--- a/encoding.bs
+++ b/encoding.bs
@@ -179,8 +179,8 @@ happening.
 <a for=/>error mode</a> <var>mode</var>, run these steps:
 
 <ol>
- <li><p>If <var>mode</var> is not given, set it to "<code>replacement</code>", if
- <var>encoderDecoder</var> is a <a for=/>decoder</a>, and "<code>fatal</code>" otherwise.
+ <li><p>If <var>mode</var> is not given, then set it to "<code>replacement</code>" if
+ <var>encoderDecoder</var> is a <a for=/>decoder</a>, otherwise "<code>fatal</code>".
 
  <li><p>Let <var>encoderDecoderInstance</var> be a new <var>encoderDecoder</var>.
 
@@ -208,9 +208,9 @@ happening.
 <a for=/>error mode</a> <var>mode</var>, run these steps:
 
 <ol>
- <li><p>If <var>mode</var> is not given, set it to "<code>replacement</code>", if
- <var>encoderDecoderInstance</var> is a <a for=/>decoder</a> instance, and "<code>fatal</code>"
- otherwise.
+ <li><p>If <var>mode</var> is not given, then set it to "<code>replacement</code>" if
+ <var>encoderDecoderInstance</var> is a <a for=/>decoder</a> instance, otherwise
+ "<code>fatal</code>".
 
  <li><p>Assert: if <var>encoderDecoderInstance</var> is an <a for=/>encoder</a> instance,
  <var>token</var> is not a <a>surrogate</a>.
@@ -278,9 +278,9 @@ from a string <var>label</var>, run these steps:
  <li><p>Remove any leading and trailing <a>ASCII whitespace</a> from
  <var>label</var>.
 
- <li><p>If <var>label</var> is an <a>ASCII case-insensitive</a>
- match for any of the <a>labels</a> listed in the table
- below, return the corresponding <a for=/>encoding</a>, and failure otherwise.
+ <li><p>If <var>label</var> is an <a>ASCII case-insensitive</a> match for any of the <a>labels</a>
+ listed in the table below, then return the corresponding <a for=/>encoding</a>; otherwise return
+ failure.
 </ol>
 
 <p class="note no-backref">This is a more basic and restrictive algorithm of mapping <a>labels</a>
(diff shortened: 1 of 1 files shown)


annevk:
@ms2ger @domenic wanna skim this?

encoding.bs
```diff
@@ -1701,6 +1706,7 @@ textReadable
 <p>The
 <dfn constructor for=TextEncoderStream id=dom-textencoderstream><code>TextEncoderStream()</code></dfn>
 constructor, when invoked, must run these steps:
+<!-- TODO: figure out constructors -->
```
domenic: So for example, this should get rid of the first and last step, and then do s/enc/this throughout.

encoding.bs
```diff
@@ -1332,6 +1335,7 @@ requires buffering of scalar values.
 
 <p>The <dfn constructor for=TextEncoder id=dom-textencoder><code>TextEncoder()</code></dfn>
 constructor, when invoked, must return a new {{TextEncoder}} object.
+<!-- TODO: what should this say in the new world? -->
```
domenic: Interesting. TextEncoder has no state so there's nothing to set up here. I guess this points to the fact that it really shouldn't be a class (it's just behavior, not data + behavior with individual instances having some meaning). But it is, presumably for symmetry, so here we are.

I guess "must do nothing"?

annevk: Do we still want to say `TextEncoder()` as a heading? Or should that also change somehow? (It could be considered a class that has the UTF-8 encoder as internal thing btw. But the spec grabs that as a global resource.)

Ms2ger:
lgtm

annevk:
@domenic @Ms2ger this is ready for another look.

encoding.bs
```diff
@@ -2725,8 +2702,8 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
 
      <li><p>Set the <a>ISO-2022-JP output flag</a>.
```
Ms2ger: Are you planning to remove these flags in a followup?

encoding.bs
```diff
@@ -1331,7 +1324,7 @@ requires buffering of scalar values.
 </dl>
 
 <p>The <dfn constructor for=TextEncoder id=dom-textencoder><code>TextEncoder()</code></dfn>
-constructor, when invoked, must return a new {{TextEncoder}} object.
+constructor, when invoked, must do nothing.
```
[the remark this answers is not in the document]
annevk: I think per @domenic it's that we use it for methods (and I was assuming constructors), but not getters/setters. Maybe I got it wrong though.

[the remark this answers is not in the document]
domenic: I think I mainly thought "on getting, when invoked" was weird. Otherwise I was probably just going for local consistency. If we wanted a global pattern we should maybe discuss.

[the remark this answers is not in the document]
annevk: It's been some time, thoughts on a global pattern?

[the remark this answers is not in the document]
domenic: Globally I think it's fairly redundant.

The overall tracking issue is https://github.com/heycam/webidl/issues/730

andreubotella:
Given the refactoring nature of this PR, I think some other minor editorial changes, like #154, could be incorporated as well, right?

encoding.bs
```diff
@@ -1110,16 +1110,20 @@ interface mixin TextDecoderCommon {
...
 <p>The {{TextDecoderCommon}} interface mixin defines common attributes that are shared between
-{{TextDecoder}} and {{TextDecoderStream}} objects. These objects have an associated
-<dfn id=textdecoder-encoding for=TextDecoderCommon>encoding</dfn>,
-<dfn id=textdecoder-ignore-bom-flag for=TextDecoderCommon>ignore BOM</dfn> (initially false),
-<dfn id=textdecoder-bom-seen-flag for=TextDecoderCommon>BOM seen</dfn> (initially false), and
-<dfn id=textdecoder-error-mode for=TextDecoderCommon>error mode</dfn> (initially
-"<code>replacement</code>").
+{{TextDecoder}} and {{TextDecoderStream}} objects. These objects have an associated:
+
+<ul class=brief>
```
domenic, line 1115: This really feels like a definition list to me, but I'm not sure if that'd end up looking right.

encoding.bs
```diff
@@ -1158,17 +1164,16 @@ control.
 <hr>
 
 <p>The <dfn attribute id=dom-textdecoder-encoding for=TextDecoderCommon><code>encoding</code></dfn>
-attribute's getter, when invoked, must return this object's <a for=TextDecoderCommon>encoding</a>'s
-<a for=encoding>name</a> in <a>ASCII lowercase</a>.
+attribute's getter steps are to return <a>this</a>'s <a for=TextDecoderCommon>encoding</a>'s
```
domenic, line 1167: Remove "attribute" per latest discussions.

encoding.bs
```diff
@@ -1236,42 +1246,38 @@ string += decoder.decode(); // end-of-stream</code></pre>
...
 
  <li><p>If <var>encoding</var> is failure or <a>replacement</a>, then <a>throw</a> a {{RangeError}}.
 
- <li><p>Let <var>dec</var> be a new {{TextDecoder}} object.
-
- <li><p>Set <var>dec</var>'s <a for=TextDecoderCommon>encoding</a> to <var>encoding</var>.
+ <li><p>Set <a>this</a>'s <a for=TextDecoderCommon>encoding</a> to <var>encoding</var>.
 
- <li><p>If <var>options</var>'s <code>fatal</code> member is true, then set <var>dec</var>'s
+ <li><p>If <var>options</var>["<code>fatal</code>"] is true, then set <a>this</a>'s
```
domenic, line 1259: You can cross-link these with `{{TextDecoderOptions/fatal}}`. Same for all other dictionary members. Maybe you'll want to delay that to a follow-up, though.

encoding.bs
```diff
@@ -1369,17 +1376,17 @@ requires buffering of scalar values.
 
  <dt><code><var>encoder</var> . <a method=for=TextEncoder lt="encodeInto(source, destination)">encodeInto(<var>source</var>, <var>destination</var>)</a></code>
  <dd><p>Runs the <a>UTF-8 encoder</a> on <var>source</var>, stores the result of that operation into
- <var>destination</var>, and returns the progress made as a dictionary whereby
+ <var>destination</var>, and returns the progress made as an object whereby
```
domenic, line 1379:
```suggestion
 <var>destination</var>, and returns the progress made as an object wherein
```

encoding.bs
```diff
@@ -1236,42 +1246,38 @@ string += decoder.decode(); // end-of-stream</code></pre>
...
-method, when invoked, must run these steps:
+method steps are:
 
 <ol>
- <li><p>If <a for=TextDecoder>do not flush</a> is false, set <a for=TextDecoder>decoder</a>
- to a new <a for=TextDecoderCommon>encoding</a>'s <a for=/>decoder</a>,
- <a for=TextDecoder>stream</a> to a new <a for=/>stream</a>, and
- <a for=TextDecoderCommon>BOM seen</a> to false.
+ <li><p>If <a>this</a>'s <a for=TextDecoder>do not flush</a> is false, then set <a>this</a>'s
+ <a for=TextDecoder>decoder</a> to a new <a for=TextDecoderCommon>encoding</a>'s
```
domenic, line 1271: This for is wrong (preexisting problem)

annevk:
@andreubotella I was going to leave that as a follow-up as changing algorithms is always somewhat risky.

encoding.bs
annevk: What is wrong about it?

domenic: The for=TextDecoderCommon should be for=/

annevk: Why?

annevk: I'm not sure how that would make sense.

domenic: You're right, that wouldn't work. Instead this should match the `new TextDecoderStream` step 6: "Set this’s decoder to a new decoder for this’s encoding"

encoding.bs
```diff
@@ -1109,17 +1109,25 @@ interface mixin TextDecoderCommon {
...
 
-<p>The {{TextDecoderCommon}} interface mixin defines common attributes that are shared between
-{{TextDecoder}} and {{TextDecoderStream}} objects. These objects have an associated
-<dfn id=textdecoder-encoding for=TextDecoderCommon>encoding</dfn>,
-<dfn id=textdecoder-ignore-bom-flag for=TextDecoderCommon>ignore BOM</dfn> (initially false),
-<dfn id=textdecoder-bom-seen-flag for=TextDecoderCommon>BOM seen</dfn> (initially false), and
-<dfn id=textdecoder-error-mode for=TextDecoderCommon>error mode</dfn> (initially
-"<code>replacement</code>").
+<p>The {{TextDecoderCommon}} interface mixin defines common getters that are shared between
+{{TextDecoder}} and {{TextDecoderStream}} objects. These objects have an associated:
```
domenic, line 1113: Although I don't have a concrete idea as to why, "an associated" followed by a `<dl>` seems weird. May I suggest "have the following associated data" or something like that? I dunno.

annevk: I'm going to leave it for now until I hear more complaints. I think it's only weird because we normally use them for switches and perhaps not always say so explicitly in the paragraph preceding one.

PR #177, opened 2019-06-17
Review Draft Publication: June 2019

annevk:
A Review Draft for this Workstream will be published shortly, by merging this pull request.

Under the [WHATWG IPR Policy](https://whatwg.org/ipr-policy), Participants may, within 45 days after publication of a Review Draft, exclude certain Essential Patent Claims from the Review Draft Licensing Obligations. See the [IPR Policy](https://whatwg.org/ipr-policy) for details.

### DIFF
--- a/review-drafts/2019-06.bs
+++ b/review-drafts/2019-06.bs
@@ -1,0 +1,3332 @@
+<pre class=metadata>
+Group: WHATWG
+Date: 2019-06-17
+H1: Encoding
+Shortname: encoding
+Text Macro: TWITTER encodings
+Abstract: The Encoding Standard defines encodings and their JavaScript API.
+Translation: ja https://triple-underscore.github.io/Encoding-ja.html
+Markup Shorthands: css off
+Translate IDs: dictdef-textdecoderoptions textdecoderoptions,dictdef-textdecodeoptions textdecodeoptions,index section-index
+</pre>
+
+<link rel=stylesheet href=visualization-colors.css>
+
+<pre class=link-defaults>
+spec:infra; type:dfn;
+    text:code point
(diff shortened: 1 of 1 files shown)


domenic:
This is failing, I think because of validator warnings on the encoding tables. /cc @sideshowbarker to verify that warnings about Unicode normalization forms would cause a build failure. In that case we should probably implement #178.

It appears the last successful build, 4 months ago, did not run the validator on the build output. https://travis-ci.org/whatwg/encoding/builds/491657427

sideshowbarker:
> This is failing, I think because of validator warnings on the encoding tables. /cc @sideshowbarker to verify that warnings about Unicode normalization forms would cause a build failure.

It’s failing for the warnings because in https://github.com/whatwg/whatwg.org/blob/17fbe5071c59194dec4ce9442fc3ab1749052cb8/resources.whatwg.org/build/deploy.sh#L186-L190 we explicitly give the `--Werror` argument in the checker invocation:

```bash
    if [ -z "$CHECKER_FILTER" ]; then
      ./vnu-runtime-image/bin/vnu --skip-non-html --Werror "$WEB_ROOT"
    else
      ./vnu-runtime-image/bin/vnu --skip-non-html --Werror --filterpattern "$CHECKER_FILTER" "$WEB_ROOT"
    fi
```

That causes warnings to be treated as errors; so if there are any warnings, the process exits non-zero.

But as I mentioned in https://github.com/whatwg/encoding/issues/178#issuecomment-503412109, I think we can work around this by making the checker suppress/filter out those warnings.

domenic:
https://encoding.spec.whatwg.org/review-drafts/2019-06/

Issue #178, opened 2019-06-19
Consider not running the validator on all the encoding tables?


domenic:
https://travis-ci.org/whatwg/encoding/builds/546617807 is taking a long time and giving a lot of warnings about Unicode normalization forms.

On the other hand, these tables should be valid HTML, I think, so maybe running the validator on them is the right thing to do?

/cc @sideshowbarker

sideshowbarker:
> https://travis-ci.org/whatwg/encoding/builds/546617807 is taking a long time and giving a lot of warnings about Unicode normalization forms.

I can make the checker suppress the warnings. Shall we try that?

PR #179, opened 2019-06-19
Suppress bad-Unicode HTML checker warnings

sideshowbarker:
This change adds an `.htmlcheckerfilter` file containing regular-expression patterns that causes some particular warnings to be suppressed when the https://github.com/whatwg/whatwg.org/blob/master/resources.whatwg.org/build/deploy.sh script invokes the HTML checker on the Encoding build output.

The particular warnings that get suppressed are:

* Text run is not in Unicode Normalization Form C.
* Document uses the Unicode Private Use Area(s), which should not be used in publicly exchanged documents. (Charmod C073)

Otherwise, without this change, deployment of the Encoding spec will fail, because the https://github.com/whatwg/whatwg.org/blob/master/resources.whatwg.org/build/deploy.sh script causes the HTML checker to be invoked with the --Werror argument, which causes warnings to be treated as errors, which in turn causes the HTML checker process to exit non-zero.

Fixes https://github.com/whatwg/encoding/issues/178
Relates to https://github.com/whatwg/encoding/pull/177

### DIFF
--- a/.htmlcheckerfilter
+++ b/.htmlcheckerfilter
@@ -1,0 +1,2 @@
+.*Text run is not in Unicode Normalization Form C.*
+.*Document uses the Unicode Private Use Area.*
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ local: encoding.bs
 
 deploy: encoding.bs
 	curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh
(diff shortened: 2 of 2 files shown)


.checker-filter
```diff
@@ -0,0 +1 @@
+.*Text.run.is.not.in.Unicode.Normalization.Form.C.*|.*Document.uses.the.Unicode.Private.Use.Area.*
```
domenic: Probably don't want this file anymore

sideshowbarker: > Probably don't want this [.checker-filter] file anymore

Yup, thanks — removed

PR #185, opened 2019-09-12
Editorial: require returned dictionary members

annevk:
As they are never not set. See https://bugzilla.mozilla.org/show_bug.cgi?id=1580154 for context.

***
<a href="https://whatpr.org/encoding/185.html" title="Last updated on Sep 12, 2019, 4:13 PM UTC (3924074)">Preview</a> | <a href="https://whatpr.org/encoding/185/8abcdaf...3924074.html" title="Last updated on Sep 12, 2019, 4:13 PM UTC (3924074)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -1293,8 +1293,8 @@ attribute's getter, when invoked, must return "<code>utf-8</code>".
 
 <pre class=idl>
 dictionary TextEncoderEncodeIntoResult {
-  unsigned long long read;
-  unsigned long long written;
+  required unsigned long long read;
+  required unsigned long long written;
 };
 
 [Constructor,


bzbarsky:
A required dictionary member is defined to be always present.  There are steps in https://heycam.github.io/webidl/#es-to-dictionary that ensure that invariant, but someone creating a dictionary on the IDL side (e.g. in a spec algorithm) must ensure that they provide a value for required members to produce a valid instance of the dictionary.  At least in Gecko, bindings enforce that.

Which means that `required` in a return value dictionary serves as useful documentation: this member will always be present in the return value.  You could get that from carefully reading the algorithm that creates the return value, of course, but I feel that it's nice to just be able to see it from the IDL.

(There may also be optimization opportunities for JITs here if they can know the shape/hidden-class/etc of the objects coming out of the call ahead of time, but I don't know how important that is in this specific case, and the in-practice monomorphic nature of the return value means that at worst the JIT ends up with one extra guard.)

annevk:
Yeah, I also like it for the reason that these are basically asserts on the return value.

If I take @domenic's stance to an extreme we'd only need to document the return type as being a dictionary, but not any particular dictionary. Mainly so that you can go from an Infra map to a dictionary to JavaScript. But this specification (and others) already document members, so we might as well document more details of those members.

annevk:
@domenic @inexorabletash could you elaborate further on why asserts on return values (that are also useful for codegen) are problematic?

domenic:
Sure.

I don't think we should be changing standard text to help code generators, unless the benefit is overwhelming for the additional cost and complexity. Here we are told that this helps Gecko in some way, but this doesn't meet the bar, in my opinion.

More troubling is the overloading of the "required" keyword. It has a normative, useful impact on input dictionaries. We encourage specification authors to use it in specific ways because of that impact. Saying that it should also be used for output dictionaries, for a completely different purpose---something which has no processing model impact, but instead is documentation/helpful for Gecko's code-gen---muddles the definition and usage of required dictionary members.

If we're looking for documentation that a dictionary is always used for output, we should document *that fact*, e.g. via a `// always output` or `// all members used when this is returned by encodeInto` comment. You know, documentation. We shouldn't have people indirectly infer that fact from noting that all members are required, and then having to check all the algorithms to find out oh, this isn't the usual "required as input" required, this is the "is always set by the spec" required.

If we're looking to improve Gecko's codegen, then Gecko should use Gecko-specific extended attributes.

Even if we had a lot of engines and parts of the ecosystem excited about codegen benefits, so as to meet the bar I mentioned earlier about putting the codegen hint in the spec, I again think overloading required is not the right way to introduce such hints; a [AlwaysSet] extended attribute would be better for that purpose.

inexorabletash:
Less rigorous feedback: For output-only dictionaries it would appear on every member, which seems to just add noise. For dictionaries that are used for both input and output (are there any?), `required` can't be present on optional input members even if the output member will always be present; do we then split the dictionary definitions just to capture this correctly? Maybe output dictionaries should be their own IDL type anyway?

(Not a strong objection from me, though.)

bzbarsky:
> are there any?

Yes.  For example, https://w3c.github.io/mediacapture-main/getusermedia.html#media-stream-track-interface-definition has `MediaTrackConstraints` as both an argument to `applyConstraints()` and a return value from `getConstraints()`.  Those could in theory be defined as two separate dictionaries, but that would make things more confusing, I think.

Similar for `RTCConfiguration`, which is an argument to the `RTCPeerConnection` constructor and a return value from `getConfiguration()`.

There are various other examples too if we care.

annevk:
Closing this per https://github.com/heycam/webidl/commit/8d4e7ca67bcb4de476aab0411531dc97f6d17e0d.

PR #196, opened 2020-01-18
Clarify that encoding tokens are scalar values.

### LINKED ISSUE #195: Clarify that encoding tokens are scalar values
It seems like the fact that Unicode tokens are scalar values, rather than code points, is far from clear in the spec. Other than the usage of `USVString` and related algorithms in the API section, the only mention of scalar values in the normative text is in the definition of encoding. In fact, the definition of token refers explicitly to code points, rather than scalar values.

This could actually result in a security issue if a specification weren't careful when using the encoding hooks – encoding handlers based on indices would raise an error on surrogate code points, but the UTF-8 handler would go along with it, returning a byte sequence which would fail on decoding.

I propose adding some text to the note in the hooks section informing specs that they should only invoke the encoding algorithms with streams built from a `USVString`, as well as adding an assertion in the process algorithm that, if `encoderDecoderInstance` is an encoder instance, `input` must not be a surrogate.

andreubotella:
This change changes some of the previous occurrences of the term "code point" for "scalar value", since it might not be clear to passing readers that output streams from decoding and input streams to encoding cannot contain tokens which are surrogate code points.

I haven't replaced the term "code point" in the indices section or in the handlers for the different encodings because those sections are heavy on numerical operations on code points, which aren't necessarily well-defined on scalar values. But perhaps the switch in terminology is more confusing than replacing the term throughout.

Fixes #195.

***
<a href="https://whatpr.org/encoding/196.html" title="Last updated on Jan 20, 2020, 1:19 PM UTC (bd1331c)">Preview</a> | <a href="https://whatpr.org/encoding/196/b818a7f...bd1331c.html" title="Last updated on Jan 20, 2020, 1:19 PM UTC (bd1331c)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -99,7 +99,7 @@ bitwise AND by "&amp;", and bitwise OR by "|".
 <hr>
 
 <p>A <dfn id=concept-token>token</dfn> is a piece of data, such as a <a>byte</a>
-or <a>code point</a>.
+or <a>scalar value</a>.
 
 <p>A <dfn id=concept-stream>stream</dfn> represents an ordered sequence of
 <a>tokens</a>. <dfn>End-of-stream</dfn> is a special
@@ -216,15 +216,24 @@ happening.
  <var>encoderDecoderInstance</var> is a <a for=/>decoder</a> instance, and "<code>fatal</code>"
  otherwise.
 
+ <li><p>Assert: if <var>encoderDecoderInstance</var> is an <a for=/>encoder</a> instance,
+ <var>token</var> is not a <a>surrogate</a>.
+
  <li><p>Let <var>result</var> be the result of running <var>encoderDecoderInstance</var>'s
  <a>handler</a> on <var>input</var> and <var>token</var>.
 
  <li><p>If <var>result</var> is <a>continue</a> or <a>finished</a>, return
  <var>result</var>.
 
- <li><p>Otherwise, if <var>result</var> is one or more
- <a>tokens</a>, <a>push</a>
- <var>result</var> to <var>output</var>.
+ <li>
+  <p>Otherwise, if <var>result</var> is one or more
+  <a>tokens</a>:
+
+  <ol>
+   <li><p>Assert: If <var>encoderDecoderInstance</var> is a <a for=/>decoder</a>
+   instance, <var>result</var> does not contain any <a>surrogates</a>.
+   <li><p><a>Push</a> <var>result</var> to <var>output</var>.
+  </ol>
 
  <li>
   <p>Otherwise, if <var>result</var> is <a>error</a>, switch on <var>mode</var> and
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -216,15 +216,22 @@ happening.
  <var>encoderDecoderInstance</var> is a <a for=/>decoder</a> instance, and "<code>fatal</code>"
  otherwise.
 
+ <li><p>Assert: If <var>encoderDecoderInstance</var> is an <a for=/>encoder</a> instance,
```
annevk:
```suggestion
 <li><p>Assert: if <var>encoderDecoderInstance</var> is an <a for=/>encoder</a> instance,
```

encoding.bs
```diff
@@ -216,15 +216,22 @@ happening.
...
+ <var>token</var> is not a <a>surrogate</a> code point.
```
annevk:
```suggestion
 <var>token</var> is not a <a>surrogate</a>.
```

encoding.bs
```diff
@@ -216,15 +216,22 @@ happening.
...
 
  <li><p>Otherwise, if <var>result</var> is one or more
- <a>tokens</a>, <a>push</a>
- <var>result</var> to <var>output</var>.
+ <a>tokens</a>:
+  <ol>
+   <li><p>Assert: If <var>encoderDecoderInstance</var> is a <a for=/>decoder</a>
+   instance, <var>result</var> does not contain any <a>surrogate</a> code points.
+   <li><p><a>Push</a> <var>result</var> to <var>output</var>.
+  </ol>
```
annevk: Can you format this as the following list item? In particular, now the list item contains multiple children, the `<p>` needs to be on its own line. Also it should be "surrogates" (no code points, as that's part of its definition).

encoding.bs
```diff
@@ -858,6 +865,10 @@ different format here, to be able to represent ranges.)
 
  <p>The <a>get an encoding</a> algorithm is to be used to turn a <a>label</a> into an
  <a for=/>encoding</a>.
+
+ <p>Standards are to ensure that the streams they pass to the <a for=/>encode</a> and
+ <a>UTF-8 encode</a> algorithms are effectively scalar value streams, i.e. they contain
```
annevk:
```suggestion
 <a>UTF-8 encode</a> algorithms are effectively scalar value streams, i.e., they contain
```

encoding.bs
```diff
@@ -858,6 +865,10 @@ different format here, to be able to represent ranges.)
...
+ no <a>surrogate</a> code points.
```
annevk:
```suggestion
 no <a>surrogates</a>.
```

andreubotella:
Is it okay then to keep using "code point" in the indices and handlers?

annevk:
Yeah, they are interchangeable.

annevk:
Thanks again @andreubotella for taking the time to write this PR!

PR #197, opened 2020-02-26
Label single-byte visualization rows by byte value

hsivonen:
As a frequent user of the visualizations, I've come to the conclusion that, for practical purposes, I always need to look up byte values rather than index offsets in the single-byte visualizations. This changes the row labels for the single-byte visualizations to use byte values.


annevk:
Thanks, this does look better (checked iso-8859-2).

PR #198, opened 2020-02-26
Editorial: change the remaining flags to booleans

### LINKED ISSUE #176: Editorial: change remaining flags to booleans
As mentioned at https://github.com/whatwg/encoding/pull/175#discussion_r287275701 this is indeed something that should be done.

The basic process here is:

* Add the ID of the flag `<dfn>` explicitly so that it doesn't change.
* Remove the word "flag" from the `<dfn>` and adjust surrounding wording accordingly to make sense for booleans.
* Ensure "flag" is no longer in the document or only in the non-boolean sense.

andreubotella:
Fixes #176.

***
<a href="https://whatpr.org/encoding/198.html" title="Last updated on Mar 2, 2020, 6:17 PM UTC (0f9aac0)">Preview</a> | <a href="https://whatpr.org/encoding/198/673e65a...0f9aac0.html" title="Last updated on Mar 2, 2020, 6:17 PM UTC (0f9aac0)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -879,7 +879,7 @@ fallback encoding <var>encoding</var>, run these steps:
 <ol>
  <li><p>Let <var>buffer</var> be an empty byte sequence.
 
- <li><p>Let <var>BOM seen</var> be false.
+ <li><p>Let <var>BOMSeen</var> be false.
 
  <li><p><a>Read</a> bytes from <var>stream</var>
  into <var>buffer</var> until either <var>buffer</var> contains three bytes or
@@ -890,7 +890,7 @@ fallback encoding <var>encoding</var>, run these steps:
   one and going down, if the first bytes of <var>buffer</var> match
   all the bytes given in the first column, then set <var>encoding</var>
   to the <a for=/>encoding</a> given in the cell in the second column of
-  that row and <var>BOM seen</var> to true.
+  that row and <var>BOMSeen</var> to true.
 
   <table>
    <tbody><tr><th>Byte order mark<th>Encoding
@@ -903,10 +903,10 @@ fallback encoding <var>encoding</var>, run these steps:
   than anything else. In a context where HTTP is used this is in violation of the semantics of the
   `<code>Content-Type</code>` header.
 
- <li><p>If <var>BOM seen</var> is false,
+ <li><p>If <var>BOMSeen</var> is false,
  <a>prepend</a> <var>buffer</var> to <var>stream</var>.
 
- <li><p>Otherwise, if <var>BOM seen</var> is true, <var>encoding</var> is not
+ <li><p>Otherwise, if <var>BOMSeen</var> is true, <var>encoding</var> is not
  <a>UTF-8</a>, and <var>buffer</var> contains three bytes,
  <a>prepend</a> the last byte of <var>buffer</var> to
  <var>stream</var>.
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -1266,12 +1264,12 @@ method, when invoked, must run these steps:
...
-    <p>If <var>token</var> is <a>end-of-stream</a> and the <a for=TextDecoder>do not flush flag</a>
-    is set, then return <var>output</var>,
+    <p>If <var>token</var> is <a>end-of-stream</a> and <a for=TextDecoder>do not flush</a>
+    is true, then return <var>output</var>,
     <a lt="serialize stream" for=TextDecoderCommon>serialized</a>.
 
-    <p class=note>The way streaming works is to not handle <a>end-of-stream</a> here when the
-    <a for=TextDecoder>do not flush flag</a> is set and to not unset that flag. That way in a
+    <p class=note>The way streaming works is to not handle <a>end-of-stream</a> here when
+    <a for=TextDecoder>do not flush</a> is true and to not set that slot to false. That way in a
```
andreubotella: The literal translation of "to not unset that flag" would be "to not set that boolean to false", but since its definition doesn't explicitly mark it as a boolean, I wasn't sure whether it was appropriate. "Slot" might not be the right term here. "Associated value" perhaps?

andreubotella:
I can't seem to be able to add reviewers for some reason, so I'm tagging @annevk here

encoding.bs
```diff
@@ -879,7 +879,7 @@ fallback encoding <var>encoding</var>, run these steps:
 <ol>
  <li><p>Let <var>buffer</var> be an empty byte sequence.
 
- <li><p>Let <var>BOM seen flag</var> be unset.
+ <li><p>Let <var>BOM seen</var> be false.
```
annevk: Can we make this `BOMSeen` while changing this? These days the convention seems to be to not use spaces for local variables.

encoding.bs
```diff
@@ -1139,8 +1139,8 @@ attribute's getter, when invoked, must return true if this object's
 
 <p>The
 <dfn attribute id=dom-textdecoder-ignorebom for=TextDecoderCommon><code>ignoreBOM</code></dfn>
-attribute's getter, when invoked, must return true if this object's
-<a for=TextDecoderCommon>ignore BOM flag</a> is set, and false otherwise.
+attribute's getter, when invoked, must return the value of this object's
```
annevk: We can also use "`<a>this</a>'s`" rather than "this object's" now.

encoding.bs
```diff
@@ -1234,14 +1233,13 @@ constructor, when invoked, must run these steps:
 method, when invoked, must run these steps:
 
 <ol>
- <li><p>If the <a for=TextDecoder>do not flush flag</a> is unset, set <a for=TextDecoder>decoder</a>
- to a new <a for=TextDecoderCommon>encoding</a>'s <a for=/>decoder</a>, set
- <a for=TextDecoder>stream</a> to a new <a for=/>stream</a>, and unset the
- <a for=TextDecoderCommon>BOM seen flag</a>.
+ <li><p>If <a for=TextDecoder>do not flush</a> is false, set <a for=TextDecoder>decoder</a>
```
annevk: Another problem with this text is that it doesn't say where to grab "do not flush" from. See #175 for an idea of how to address that. Although I guess I'd accept it if you don't want to take that on as well.

encoding.bs
annevk: s/that slot/it/ would work? Or duplicate the term.

encoding.bs
andreubotella: Both this change and updating cases of "this object" to "this" are changes that go a bit outside the scope of this pull request. Since #175 seems to be dead, maybe some similar PR, though smaller in scope, should be opened.

annevk:
Thanks @andreubotella and totally reasonable regarding scope.

PR #203, opened 2020-03-16
Add a BOM sniffing hook for better integration with HTML

### LINKED ISSUE #128: Figure out if anything is needed for better HTML integration
@hsivonen in https://github.com/whatwg/html/issues/1077#issuecomment-214780498 raised a number of issues with the current integration points. They are insufficient for CSS, HTML, and presumably XML.

This might require some substantive changes to the hooks and perhaps other parts of the Encoding Standard, as well as standards that depend on the Encoding Standard (of which there are quite a few, so tread carefully).

Belated filing this to keep better track of it.

andreubotella:
This change moves the BOM splitting part of the decode hook into a separate hook which does not consume any bytes of the token stream.

This will allow fixing a long-standing issue in the HTML encoding sniffing algorithm with the document's character encoding being set to the wrong result when there is a BOM (see whatwg/html#1077).

Closes #128.

***
<a href="https://whatpr.org/encoding/203.html" title="Last updated on Mar 24, 2020, 8:24 AM UTC (ed1ec60)">Preview</a> | <a href="https://whatpr.org/encoding/203/64e63cc...ed1ec60.html" title="Last updated on Mar 24, 2020, 8:24 AM UTC (ed1ec60)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -854,7 +854,7 @@ different format here, to be able to represent ranges.)
 <div class=note>
  <p>The algorithms defined below (<a>decode</a>, <a>UTF-8 decode</a>,
  <a>UTF-8 decode without BOM</a>, <a>UTF-8 decode without BOM or fail</a>, <a for=/>encode</a>,
- <a>UTF-8 encode</a> and <a>BOM sniff</a>) are intended for usage by other standards.
+ <a>UTF-8 encode</a>, and <a>BOM sniff</a>) are intended for usage by other standards.
 
  <p>For decoding, <a>UTF-8 decode</a> is to be used by new formats. For identifiers or byte
  sequences within a format or protocol, use <a>UTF-8 decode without BOM</a> or
@@ -862,8 +862,8 @@ different format here, to be able to represent ranges.)
 
  <p>For encoding, <a>UTF-8 encode</a> is to be used.
 
- <p>Standards are strongly discouraged from using <a>decode</a>, <a for=/>encode</a> and <a>BOM
- sniff</a>, except as needed for compatibility.
+ <p>Standards are strongly discouraged from using <a>decode</a>, <a for=/>encode</a>, and
+ <a>BOM sniff</a>, except as needed for compatibility.
 
  <p>The <a>get an encoding</a> algorithm is to be used to turn a <a>label</a> into an
  <a for=/>encoding</a>.
@@ -877,8 +877,10 @@ different format here, to be able to represent ranges.)
 fallback encoding <var>encoding</var>, run these steps:
 
 <ol>
- <li><p><a>BOM sniff</a> <var>stream</var>. If the result is not null, set <var>encoding</var> to
- the resulting value and let <var>BOMSeen</var> be true. Otherwise, let <var>BOMSeen</var> be false.
+ <li>
+  <p><a>BOM sniff</a> <var>stream</var>. If the result is not null, set <var>encoding</var> to
+  the resulting value and let <var>BOMSeen</var> be true. Otherwise, let <var>BOMSeen</var> be
+  false.
 
   <p class=note>For compatibility with deployed content, the byte order mark is more authoritative
   than anything else. In a context where HTTP is used this is in violation of the semantics of the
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -853,17 +853,17 @@ different format here, to be able to represent ranges.)
 
 <div class=note>
  <p>The algorithms defined below (<a>decode</a>, <a>UTF-8 decode</a>,
- <a>UTF-8 decode without BOM</a>, <a>UTF-8 decode without BOM or fail</a>, <a for=/>encode</a>, and
- <a>UTF-8 encode</a>) are intended for usage by other standards.
+ <a>UTF-8 decode without BOM</a>, <a>UTF-8 decode without BOM or fail</a>, <a for=/>encode</a>,
+ <a>UTF-8 encode</a> and <a>BOM sniff</a>) are intended for usage by other standards.
```
annevk:
```suggestion
 <a>UTF-8 encode</a>, and <a>BOM sniff</a>) are intended for usage by other standards.
```

encoding.bs
```diff
@@ -853,17 +853,17 @@ different format here, to be able to represent ranges.)
...
  <p>For decoding, <a>UTF-8 decode</a> is to be used by new formats. For identifiers or byte
  sequences within a format or protocol, use <a>UTF-8 decode without BOM</a> or
  <a>UTF-8 decode without BOM or fail</a>.
 
  <p>For encoding, <a>UTF-8 encode</a> is to be used.
 
- <p>Standards are strongly discouraged from using <a>decode</a> and <a for=/>encode</a>, except as
- needed for compatibility.
+ <p>Standards are strongly discouraged from using <a>decode</a>, <a for=/>encode</a> and <a>BOM
+ sniff</a>, except as needed for compatibility.
```
annevk:
```suggestion
 <p>Standards are strongly discouraged from using <a>decode</a>, <a for=/>encode</a>, and
 <a>BOM sniff</a>, except as needed for compatibility.
```

encoding.bs
```diff
@@ -877,39 +877,15 @@ different format here, to be able to represent ranges.)
...
-  to the <a for=/>encoding</a> given in the cell in the second column of
-  that row and <var>BOMSeen</var> to true.
-
-  <table>
-   <tbody><tr><th>Byte order mark<th>Encoding
-   <tr><td>0xEF 0xBB 0xBF<td><a>UTF-8</a>
-   <tr><td>0xFE 0xFF<td><a>UTF-16BE</a>
-   <tr><td>0xFF 0xFE<td><a>UTF-16LE</a>
-  </table>
+ <li><p><a>BOM sniff</a> <var>stream</var>. If the result is not null, set <var>encoding</var> to
```
annevk: Let's make this something like  Let `BOMEncoding` be the result of BOM sniffing stream and then drop the BOMSeen variable. And override `encoding` in a subsequent step.

annevk: Also, the p will need to be on its own line as the li has two children.

encoding.bs
```diff
@@ -995,7 +971,34 @@ steps:
 <p>To <dfn export>UTF-8 encode</dfn> a scalar value stream <var>stream</var>, return the result of
 <a lt=encode for=/>encoding</a> <var>stream</var> using encoding <a>UTF-8</a>.
 
+<hr>
+
+<p>To <dfn export>BOM sniff</dfn> a byte stream <var>stream</var>, run these steps:
+
+<ol>
+  <li><p>Wait until <var>stream</var> has three bytes available or the <a>end-of-stream</a> has been
```
annevk: One space for markup indentation here and below. (And more indentation for the children of the table, to match what it was before.

encoding.bs
```diff
@@ -995,7 +971,34 @@ steps:
...
+    <tr><td>0xFF 0xFE<td><a>UTF-16LE</a>
+    </table>
+
+  <li><p>Otherwise, return null.
+</ol>
 
+<p class=note>This hook is a workaround for the fact that <a>decode</a> has no way to communicate
+back to the caller that it has found a byte order mark and is therefore not using the provided
+encoding. The hook is to be invoked before <a>decode</a>, and it will return an encoding
+corresponding to the byte order mark found, or null otherwise.
```
annevk: You need to add some newlines here to preserve the original number of newlines before a heading.

andreubotella: There are two newlines here in the original, but elsewhere in the spec there are three newlines before an `<h2>`.

annevk: I count three unchanged blank lines in the diff.

andreubotella: My bad, I was doing `git diff` with respect to origin/master and it was matching one of the blank lines in the original in the middle of my change.

andreubotella:
Yeah, that works.

PR #210, opened 2020-05-06
Meta: adopt Python 3

annevk:


tools-index.py
```diff
@@ -36,7 +30,7 @@ def get_name(cp):
     elif cp >= 0xAC00 and cp <= 0xD7A3:
         #return "<Hangul Syllable>"
         i = cp - 0xAC00
-        s = jamo[0][i/28/21] + jamo[1][i/28%21] + jamo[2][i%28]
+        s = jamo[0][int(i/28/21)] + jamo[1][int(i/28%21)] + jamo[2][i%28]
```
hsivonen, line 33: Please use the `//` integer division operator instead of using `int()` afterwards.

PR #215, opened 2020-05-28
Rename Encoding's "streams" to "I/O queues"

### LINKED ISSUE #180: Export `pushed` concept to ReSpec
For the [`pushed`](https://encoding.spec.whatwg.org/#concept-stream-push) concept, which is not exportable and can not be searched from [ReSpec UI](https://respec.org/xref/).

Could someone help make it exportable? Probably just adding `data-export=""` to the `<dfn>`.

CC / @marcoscaceres

andreubotella:
This change renames the Encoding-specific concept of "streams", which had been causing confusion with readable/writable streams for years, to ~"token queues"~ "I/O queues". It also exports the corresponding definitions.

Closes #180.

***
<a href="https://whatpr.org/encoding/215.html" title="Last updated on Sep 8, 2020, 10:34 AM UTC (fa4a2cf)">Preview</a> | <a href="https://whatpr.org/encoding/215/e11e447...fa4a2cf.html" title="Last updated on Sep 8, 2020, 10:34 AM UTC (fa4a2cf)">Diff</a>


annevk:
I realize the timing of this is not great, but looking at this I wonder if this should be (a subtype of) https://infra.spec.whatwg.org/#queues. The main novelty is returning end-of-stream (which we should rename to final-item or final-token, I think) when the list is empty. And even that seems handled in a way as Infra returns nothing which is something we could branch on.

At that point all that remains is mapping strings/byte sequences to lists which is something we should allow implicitly anyway I think so "for each" and such can be used on them (although for strings we might need an explicit variant for code points; if you want neither code units nor scalar values).

andreubotella:
> I realize the timing of this is not great, but looking at this I wonder if this should be (a subtype of) https://infra.spec.whatwg.org/#queues. The main novelty is returning end-of-stream (which we should rename to final-item or final-token, I think) when the list is empty. And even that seems handled in a way as Infra returns nothing which is something we could branch on.

I don't think token queues are a subset of Infra queues, since prepend is a thing. Also, since read would need changes from dequeue, and you can push multiple tokens at a time which you can't do with enqueue, no real benefit would come from depending on queue. Making token queues a subset of list, rather than them being an "ordered sequence", would work though.

By the way, should we even export prepend? The "implementation considerations" appendix lists alternatives to implementing prepend which work for the encoding algorithms in the spec, but wouldn't if other specs are allowed to use prepend arbitrarily.

> At that point all that remains is mapping strings/byte sequences to lists which is something we should allow implicitly anyway I think so "for each" and such can be used on them (although for strings we might need an explicit variant for code points; if you want neither code units nor scalar values).

While I don't oppose defining strings and byte sequences as lists, I don't see how token streams would benefit from being able to iterate through them without dequeuing tokens, which is what is usually intended.

In any case, the fact that token queues can be implicitly converted to and from strings/byte sequences should be specified. Which brings me to wondering whether the conversion into a string/byte sequence should indeed empty the queue, since if the token queue is backed by I/O it'd have to block either way. If that is the case, then the BOM sniff hook would have to switch to read and prepend rather than use "starts with".

> (What it would continue to hide/neglect, which may or may not be bad, is some kind of waiting signal to indicate the difference between the end and I/O being slow.)

Token queues are defined as simple list-like data structures, not dependent on I/O, which implies that a straightforward implementation would have to read a byte stream from the network in its entirety before passing it to one of the decode hooks, with the only affordance for I/O in actual implementations is BOM sniff's (formerly decode's) "wait for three bytes or until the end-of-stream". So doing something like that would require changing token queues to optionally be backed by I/O, which would need a separate PR.

annevk:
Thanks. Given your feedback it seems somewhat tempting to try to remove the need for prepend and add the ability to enqueue multiple items to queues. If we go with a list subtype I think we need a different name from queues. Maybe consumable list or some such?

> If that is the case, then the BOM sniff hook would have to switch to read and prepend rather than use "starts with".

Yeah, I think that would be better. Even if we allow conversions without explicit casting it seems best for these algorithms to operate as if they are getting an I/O stream.

andreubotella:
As I mentioned above, it bugged me a bit the fact that token queues were defined as simple data structures but were in practice used for I/O. I've now come up with a relatively simple way to solve this problem, and so I've opened #221 with that proposal. Please let me know what you think.

annevk:
I think that's reasonable, but I do think we need a different name from queue as it's to confusing with the Infra primitives.

andreubotella:
>I think that's reasonable, but I do think we need a different name from queue as it's too confusing with the Infra primitives.
So "streams" as defined here are properly a deque, for which we don't have any Infra primitives. It'd be okay by me to define it as a list, though IMO "constructable list" sounds too generic – I'd prefer something specific to tokens or encoding.

annevk:
Hmm, so:

1. I think per our IRC conversation it's okay for this to be fully blocking and synchronous. If you need things to be non-blocking and asynchronous, use "in parallel".
2. We're likely not going to rewrite the Encoding standard to not need "prepend" so we need something like list.
3. As currently defined in the Encoding standard, "token" is just another word for Infra's "item". Token being a byte or scalar value are examples, it's not exhaustive.

I think rewriting everything in terms of lists would probably add clarity. I guess the one change we could make is that EOF should be explicitly be pushed into the list so we can "wait" for the list to change when it's empty.

We might also want to define some shorthands for byte list and scalar value list or some such that clarify the types involved.

andreubotella:
> I think rewriting everything in terms of lists would probably add clarity. I guess the one change we could make is that EOF should be explicitly be pushed into the list so we can "wait" for the list to change when it's empty.

@annevk So if I'm understanding this right, reading would mean waiting until there is at least one item in the list; and if that item is EOF, return an EOF without popping that one off the list.

Also, we should probably add a "new consumable list" operation which populates it with the EOF, and a note for users of the spec and implementers on how to create IO-backed consumable lists.

> 1. I think per our IRC conversation it's okay for this to be fully blocking and synchronous. If you need things to be non-blocking and asynchronous, use "in parallel".

Come to think about it, the encoding APIs can't be made async for web compat, which precludes running them in parallel, and aren't allowed to block because they must be callable from window agents. The way the spec text is right now, the blocking is hidden behind the scenes, but the APIs don't run into this issue because none of the lists used in those algorithms are IO-backed. So we should not make "read" always potentially blocking, but instead have it depend on whether the list is IO-backed -- or alternative, on whether the list was created with an EOF.

annevk:
Either that or you ensure the API callers always supply a list that contains EOF so wait won't be used, right?

I'm not sure I understand not needing wait.

andreubotella:
> I'm not sure I understand not needing wait.

Wait is only used so that the encoding and MIME sniffing algorithms can work on a delimited prefix of the consumable list as if it were a byte sequence. But IIRC we discussed previously that the [conversions from a consumable list into a string or byte sequence](https://whatpr.org/encoding/215.html#ref-for-concept-stream%E2%91%A4) (implicit until this PR) would consume the entire list and block until an EOF was found, making "wait" useless. Instead, we should have a peek operation that would take a length, block until the list has that many items or an EOF, and return the prefix of the consumable list as a string or byte sequence.

But if strings and byte sequences are going to be defined as subtypes of list at some point in the future, maybe it'd be best to scrap those conversions and continue using consumable lists as strings and byte sequences.

andreubotella:
Per an IRC discussion, I'm renaming streams / token queues to I/O queues.

With regard to I/O queue being (by the conversations earlier in this thread) a subtype of list, @domenic pointed to HTML's [task queue](https://html.spec.whatwg.org/multipage/webappapis.html#task-queue), which is defined as a subtype of set. I additionally pointed that the "prepend" algorithm (the only thing that makes I/O queue not a subtype of queue) is an internal implementation detail of the spec, as evidenced by the "Implementation considerations" section and by the decision earlier in this thread that we should not export that operation.

encoding.bs
```diff
@@ -94,37 +96,177 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+values</a>). <dfn id="end-of-stream" export>End-of-file</dfn> is a special
+<a>item</a> that can be present in <a for=/>I/O queues</a> of any type, and that
+signifies no more <a>items</a> are in the queue.
 
-<p>A <dfn id=concept-stream>stream</dfn> represents an ordered sequence of
-<a>tokens</a>. <dfn>End-of-stream</dfn> is a special
-<a>token</a> that signifies no more
-<a>tokens</a> are in the
-<a for=/>stream</a>.
+<p>To <dfn>create an I/O queue</dfn>, return a new <a for=/>I/O queue</a>
```
annevk, line 105: Shouldn't this also take some data as an argument to put in the queue?

encoding.bs
```diff
@@ -94,37 +96,177 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+ <a for=list>size</a> is at least 1.
+
+ <li><p>If the first item in <var>ioQueue</var> is <a>end-of-file</a>, then
+ return <a>end-of-file</a>.
+
+ <li><p>Otherwise, <a for=list>remove</a> the first item in <var>ioQueue</var>
+ and return it.
+</ol>
+
+<p>To read a number <var>number</var> of <a>items</a> from <var>ioQueue</var>,
```
annevk, line 141: If we need this primitive it should have a dfn, no?

encoding.bs
```diff
@@ -94,37 +96,177 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+algorithms in this specification, and is not to be used by other standards.
+Implementations are free to find alternative ways to implement such algorithms,
+as detailed in [[#implementation-considerations]].
+
+<hr>
+
+<p><a for=/>I/O queues</a> can be implicitly converted into and from
+<a>strings</a> and <a>byte sequences</a> using the following operations.
+
+<p>To <dfn for="byte sequence to I/O queue" export>convert</dfn> a <a>byte
```
annevk, line 198: If we're going to define this as part of this PR we should also reference it where it's needed below I think.

encoding.bs
```diff
@@ -94,37 +96,177 @@ bitwise AND by "&amp;", and bitwise OR by "|".
 
 <hr>
 
-<p>A <dfn id=concept-token>token</dfn> is a piece of data, such as a <a>byte</a> or
-<a>scalar value</a>.
+<p>An <dfn id=concept-stream export>I/O queue</dfn> is a type of <a>list</a>
+with <a>items</a> of a particular type (i.e. <a>bytes</a> or <a>scalar
+values</a>). <dfn id="end-of-stream" export>End-of-file</dfn> is a special
```
annevk, line 101: end-of-queue?

encoding.bs
```diff
@@ -13,6 +13,8 @@ Translate IDs: dictdef-textdecoderoptions textdecoderoptions,dictdef-textdecodeo
 
 <pre class=link-defaults>
 spec:streams; type:interface; text:ReadableStream
+spec:infra; type:dfn; text:list;
```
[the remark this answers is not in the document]
andreubotella: What about "item"? I'm using it to replace "token", so that's a lot of usages where I'd have to use `<a for=list>item</a>`.

encoding.bs
andreubotella: Here I was trying to take algorithms defined to take "one or more tokens" and split them into two algorithms that took an item and a sequence. But maybe I should try and merge both overloads into one algorithm.

encoding.bs
andreubotella: Here I was trying to take algorithms defined to take "one or more tokens" and split them into two algorithms that took an item and a sequence. But maybe I should try and merge both overloads into one algorithm.

encoding.bs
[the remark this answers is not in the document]
annevk: I currently only have 4 hits for `<a>token</a>` so that should be fine?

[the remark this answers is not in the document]
andreubotella: There are a few more uses introduced in this PR, but you're right, they're not that many.

encoding.bs
andreubotella: I was thinking of "create an I/O queue" as an operation that would replace the "a new stream" text in the API algorithms: it creates an "empty" immediate queue and leaves it to the calling algorithm to push items into it, sincepush is careful to sidestep the end-of-queue item. But if you think it should take data to initialize the queue, that's fine.

encoding.bs
```diff
@@ -152,6 +152,28 @@ run these steps:
  <li><p>Return <var>readItems</var>.
 </ol>
 
+<p>To <dfn for="I/O queue" export>peek</dfn> a number <var>number</var> of <a for=list>items</a>
```
andreubotella, line 155: I keep going back and forth on whether we should keep the conversion algorithms and add a peek operation, or remove the conversions and add wait. See https://github.com/whatwg/encoding/pull/215#issuecomment-658836516

encoding.bs
```diff
@@ -102,17 +102,72 @@ values</a>). <dfn id="end-of-stream" export>End-of-file</dfn> is a special
...
-     SimonSapin thinks this is fine, blame him if not -->
+<p>To <dfn>create an I/O queue</dfn>, return a new <a for=/>I/O queue</a>
+containing an <a>end-of-file</a> item.
+
+<div class=note>
+ <p>There are two ways to use an <a for=/>I/O queue</a>: in immediate
+ mode, where the queue's last item is <a>end-of-file</a>, and where the
+ <a for="I/O queue">read</a> operation cannot block; and in streaming mode, to
+ represent data coming in from the network, where the queue need not have an
+ <a>end-of-file</a> item and <a for="I/O queue">read</a> might block.
```
andreubotella: Should we move some of this into normative text and define "immediate mode" and "streaming mode" as dfn's?

encoding.bs
annevk: I don't think we necessarily need to add conversions now. And we probably want to do something in Infra around conversions of these I suspect as they're generally useful.

encoding.bs
annevk: I would expect I/O queue to have a non-blocking I/O queue subtype or some such that always contains an end-of-file, but I don't think we should do that from the start.

encoding.bs
annevk: It might be simpler to write « end-of-file » if you want that.

encoding.bs
andreubotella: There are algorithms that use those conversions, and that before this PR weren't linked to any dfn. We might want to have a short paragraph defining how conversion works with consuming the queue and creating a queue with an EOF, and have a `class=XXX` note saying that Infra is expected to define infrastructure around conversions.

encoding.bs
```diff
@@ -94,35 +90,133 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+<p>An <dfn id=concept-stream export>I/O queue</dfn> is a type of <a for=/>list</a>
+with <a for=list>items</a> of a particular type (i.e. <a>bytes</a> or <a>scalar
+values</a>). <dfn id="end-of-stream" export>End-of-queue</dfn> is a special
+<a for=list>item</a> that can be present in <a for=/>I/O queues</a> of any type, and that
+signifies no more <a for=list>items</a> are in the queue.
+
+<p>To <dfn for="I/O queue" export>create</dfn> an <a for=/>I/O queue</a>, optionally initialized
+with a list <var>contents</var>, return a new <a for=/>I/O queue</a> containing the
+<a for=list>items</a> of <var>contents</var> in order, if it is given, followed by an
+<a>end-of-queue</a> item.
```
annevk, line 102: Do we really need the create primitive instead of using the shorthand syntax for lists? It seems confusing to me to have it and only allow immediate-queue creation.

encoding.bs
```diff
@@ -94,35 +90,133 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+</div>
+
+<p>To <dfn id=concept-stream-read for="I/O queue" export>read</dfn> an
+<a for=list>item</a> from an <a for=/>I/O queue</a> <var>ioQueue</var>, run these steps:
+
+<ol>
+ <li><p>If <var>ioQueue</var> is <a for=list>empty</a>, then wait until its
+ <a for=list>size</a> is at least 1.
+
+ <li><p>If the first item in <var>ioQueue</var> is <a>end-of-queue</a>, then
```
annevk, line 130: ioQueue[0] is ...

encoding.bs
```diff
@@ -94,35 +90,133 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+ return <a>end-of-queue</a>.
+
+ <li><p>Otherwise, <a for=list>remove</a> the first item in <var>ioQueue</var>
+ and return it.
+</ol>
+
+<p>To read a number <var>number</var> of <a for=list>items</a> from <var>ioQueue</var>,
```
annevk, line 137: This needs a dfn if we are to keep it.

encoding.bs
andreubotella: I was under the impression that the shorthand syntax couldn't be used for list subtypes, but I just checked and saw an example being used for sets, so I guess it's fine.

encoding.bs
andreubotella: Should I use something like `<dfn lt="read" for="I/O queue">read an item</dfn>` vs `<dfn for="I/O queue">read a number of items</dfn>`?

annevk: Thinking about this some more maybe overloading is the way to go. In that case we should xref "read" here and leave it at that I think. (I tried thinking of some alternative ways of doing this including "read"/"read multiple" (looks awkward) or "read one or more" (but then return type is annoying), and I can't really figure it out.

Perhaps @domenic has thoughts on this.

encoding.bs
```diff
@@ -94,35 +90,125 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
-<p>When one or more <a>tokens</a> are
-<dfn id=concept-stream-push for=stream>pushed</dfn> to a <a for=/>stream</a>,
-those tokens must be inserted, in given order, after the last token in the stream.
+ <li><p>Let <var>prefix</var> be an empty list.
+
+ <li>
+  <p><a for=list>For each</a> <var>n</var> in <a>the range</a> 1 to <var>number</var>, inclusive:
+
+  <ol>
+   <li><p>If <var>ioQueue</var>[<var>n</var>] is <a>end-of-queue</a>, <a>break</a>. Otherwise,
```
annevk, line 162: Let's make this two steps?

encoding.bs
```diff
@@ -94,35 +90,125 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+   <a for=list>append</a> <var>ioQueue</var>[<var>n</var>] to <var>prefix</var>.
+  </ol>
+ </li>
+
+ <li><p>Return <var>prefix</var>.
+</ol>
+
+<p>To <dfn id=concept-stream-push for="I/O queue" export>push</dfn> one
```
annevk, line 170: s/one/an/

encoding.bs
```diff
@@ -94,35 +90,125 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+<a for=list>item</a> <var>item</var> to an <a for=/>I/O queue</a> <var>ioQueue</var>, run
+these steps:
+
+<ol>
+ <li><p>If the last <a for=list>item</a> in <var>ioQueue</var> is <a>end-of-queue</a>,
+ then <a for=list>insert</a> <var>item</var> before the last <a for=list>item</a> in the
+ queue.
```
annevk, line 177: s/the queue/_ioQueue_/

encoding.bs
```diff
@@ -94,35 +90,125 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+
+ <li><p>Otherwise, <a for=list>append</a> <var>item</var> to <var>ioQueue</var>.
+</ol>
+
+<p>To push a sequence of items is to push each item in the sequence, in the
```
annevk, line 182: We should probably use equivalent wording to reading here.

encoding.bs
```diff
@@ -94,35 +90,125 @@ bitwise AND by "&amp;", and bitwise OR by "|".
...
+given order.
+
+<p>To <dfn id=concept-stream-prepend for="I/O queue">prepend</dfn> one
+<a for=list>item</a> to an <a for=/>I/O queue</a>, perform the normal <a for=/>list</a>
+<a for=list>prepend</a> operation. To prepend a sequence of items, insert those
+items, in the given order, before the first item in the queue.
+
+<p class=example id=example-items>Inserting the sequence of scalar value items
```
annevk, line 190: Let's not change the ID here.

encoding.bs
```diff
@@ -991,22 +1080,22 @@ steps:
...
 <hr>
 
-<p>To <dfn export>BOM sniff</dfn> a byte stream <var>stream</var>, run these steps:
+<p>To <dfn export>BOM sniff</dfn> an I/O queue of bytes <var>ioQueue</var>, run these steps:
 
 <ol>
- <li><p>Wait until <var>stream</var> has three bytes available or the <a>end-of-stream</a> has been
- reached, whichever comes first.
+ <li><p>Let <var>BOM</var> be the result of <a for="I/O queue">reading</a> 3 bytes from
+ <var>ioQueue</var>, converted to a byte sequence.
```
annevk, line 1092: xref convert?

annevk (commented):
Apologies for the slow follow-up here.

annevk:
Anyone any final thoughts on this? @domenic I think this requires your PR to be rebased. I can help with that.

If there's nothing further I'll merge this later this week.

andreubotella:
>Anyone any final thoughts on this?

While working on whatwg/html#5874, I run into the fact that the HTML parser runs on the event loop and so isn't supposed to block. While the tokenizer's "consume the next input character" could be formalized into not blocking by only reading after checking that the input stream isn't empty, the call to decode would still block until the end of the request body.

I previously discussed the problem of blocking on the HTML parser with @annevk, and they suggested changing the different parser stages to run in parallel. But for decode and the rest of hooks, that wouldn't work, since the output queue is returned from the hooks, as well as being an immediate queue. I guess we could fix this without breaking other usages of the hooks by taking the output stream as an optional parameter.

encoding.bs
```diff
@@ -999,17 +1000,16 @@ fallback encoding <var>encoding</var>, run these steps:
...
   `<code>Content-Type</code>` header.
 
- <li><p>Let <var>output</var> be the <a for=/>I/O queue</a> of scalar values
- « <a>end-of-queue</a> ».
+ <li><p><a>Run</a> <var>encoding</var>'s <a for=/>decoder</a> with <var>ioQueue</var> and
+ <var>output</var>.
 
- <li><p><a>Run</a> <var>encoding</var>'s
- <a for=/>decoder</a> with <var>ioQueue</var> and <var>output</var>.
+ <li><p><a for="I/O queue">Push</a> <a>end-of-queue</a> into <var>output</var>.
```
andreubotella, line 1006: Or should the "run" algorithm push end-of-queue?

encoding.bs
```diff
@@ -1018,61 +1018,56 @@ run these steps:
...
-
- <li><p>Let <var>potentialError</var> be the result of <a>running</a>
- <a>UTF-8</a>'s <a for=/>decoder</a> with <var>ioQueue</var>, <var>output</var>,
- and "<code>fatal</code>".
+ <li><p>Let <var>potentialError</var> be the result of <a>running</a> <a>UTF-8</a>'s
+ <a for=/>decoder</a> with <var>ioQueue</var>, <var>output</var>, and "<code>fatal</code>".
 
  <li><p>If <var>potentialError</var> is <a>error</a>, return failure.
 
+ <li><p><a for="I/O queue">Push</a> <a>end-of-queue</a> into <var>output</var>.
```
andreubotella, line 1053: Do we push in the failure case too? For the case of an XML parser with the tokenizer running in parallel from the decoding, we wouldn't want the tokenizer to process the end-of-queue token before the document is blown away, even in cases where the tokenizer wouldn't error (say, inside a comment after the root element's end tag).

andreubotella:
The message of the original commit in this PR no longer reflects the total of the changes. Instead, it should be squashed and merged with the following message:

```Rename Encoding's "streams" to "I/O queues"

This change renames the Encoding-specific concept of "streams", which
had been causing confusion with readable/writable streams for years, to
"I/O queues". It also refactors the I/O queue operations and exports the
corresponding definitions.

As part of this refactoring, "end-of-queue" (formerly "end-of-stream")
becomes an optional item in the queue, indicating that the end of the
streaming data has been reached and that no more data is expected. As a
result, the "read" operation explicitly blocks when trying to read from
an empty queue – a behavior that was previously left unstated.

Closes #180.```

encoding.bs
annevk: As I wrote on IRC, how would you feel about making "process" do it?

encoding.bs
annevk: So currently "Hooks for standards" doesn't really cover the XML case. That needs a variant of "decode" that captures the output of "run" and also passes fatal as error mode in. I can file a follow-up issue for XML if you agree.

encoding.bs
andreubotella: Works for me

encoding.bs
andreubotella: Making a hook for XML, that'd be fine by me, but I don't think it would be needed until and unless there's an algorithmic XML parser defined somewhere, which seems unlikely to happen anytime soon.

triple-underscore:
The output argument “I/O queue of scalar values” in the following algorithm description seems mistyped, which would be “I/O queue of bytes” (before [the commit](https://github.com/whatwg/encoding/commit/46711e682e285497d7efc8442c80a07da104b3f0), that argument was introduced as a byte stream variable in the encode algorithm):

> To [encode](https://encoding.spec.whatwg.org/#encode) an I/O queue of scalar values ioQueue given an encoding encoding and an optional I/O queue of scalar values output (default « »), run these steps: 

> To [UTF-8 encode](https://encoding.spec.whatwg.org/#utf-8-encode) an I/O queue of scalar values ioQueue given an optional I/O queue of scalar values output (default « »), return the result of encoding ioQueue with encoding UTF-8 and output.

andreubotella:
> The output argument “I/O queue of scalar values” in the following algorithm description seems mistyped, which would be “I/O queue of bytes” (before [the commit](https://github.com/whatwg/encoding/commit/46711e682e285497d7efc8442c80a07da104b3f0), that argument was introduced as a byte stream variable in the encode algorithm):

Damn copy-paste 😄

PR #218, opened 2020-06-11
Meta: update repository files

foolip:
See https://github.com/whatwg/spec-factory for details.


annevk:
Cool, this looks like it does the right thing.

foolip:
Success: https://github.com/whatwg/encoding/runs/769616588 / https://encoding.spec.whatwg.org/

Building is pretty slow though, over 7 minutes.

annevk:
Yeah, that's expected.There's a lot of data being generated. An alternative might be that we run it locally and check the output into the repository. It currently has a mix of both approaches.

PR #221, opened 2020-06-25
Specify the usage of token queues ("streams") for I/O

andreubotella:
For a long time, token queues (formerly "streams") had been defined as simple data structures, with no reference to I/O except for a few uses of a "wait until" operation not defined anywhere and a comment on the spec's source saying that the read operation blocks on I/O.

This change introduces the concept of "blocking tokens", which are allowed to block when being read from the token queue. This makes it possible for token queues to be used as a direct spec representation of implementation-internal I/O streams, while all other operations done on token queues use immediate tokens.

This pull request depends on #215, which introduces the name "token queues" and defines the implicit conversions with strings and byte sequences.

<hr>

Although this change isn't editorial, all implementations are in fact using token queues for I/O, but I haven't checked yet if my changes conflict with the implementations.

   * …
   * …
   * …
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/encoding/221.html" title="Last updated on Jun 25, 2020, 11:57 AM UTC (77709e0)">Preview</a> | <a href="https://whatpr.org/encoding/221/e11e447...77709e0.html" title="Last updated on Jun 25, 2020, 11:57 AM UTC (77709e0)">Diff</a>


annevk:
This seems like a good idea, but read shouldn't return the blocking token, right? It should wait and then try again, until there is a non-blocking token to return?

andreubotella:
The idea here was to make as few changes as possible to the current model, which already treats a stream with data from I/O as containing the whole content when it's created. Since the only way to tell the current length of the stream is by reading tokens until EOS, I'm treating its contents that come from I/O as spec artifacts unobservable from outside the token queue.

But perhaps this model isn't too easy to understand and it might be best to simply have one special blocking token that, when read, would prepend a token to the queue before it. That's probably far more simple and would better reflect what implementations do.

andreubotella:
After some discussion on IRC, I'm closing this PR, since there are alternatives that are far easier to understand. Additionally, it's best to not discuss this change while the new name for streams is being bikeshed in #215.

PR #227, opened 2020-09-01
Meta: cleanup visualize.py

annevk:
Not entirely sure how I stumbled upon this, but I checked if this was still relevant and it's not so I decided to patch it. It didn't seem appreciably faster or slower either, though I didn't do detailed measurements.

### DIFF
--- a/visualize.py
+++ b/visualize.py
@@ -68,7 +68,7 @@ big5_prefer_last = [
 
 def classify(code_point):
     if code_point < 0x80:
-        raise Error()
+        raise Exception()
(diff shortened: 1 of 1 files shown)


hsivonen:
Does it still work with Apple-supplied Python afterwards?

annevk:
That already broke in #210. We haven't attempted to keep Python 2 compatibility anywhere in WHATWG scripts and I don't think it's worth the effort.

hsivonen (approved):

PR #230, opened 2020-09-17
Clarify the usage the output I/O queues in the hooks

andreubotella:
An "output" parameter was added to the hooks for standards in whatwg/encoding#215, but no explanation was given as to why it was needed. This change adds that clarification.

***
<a href="https://whatpr.org/encoding/230.html" title="Last updated on Sep 21, 2020, 10:09 AM UTC (eade6ff)">Preview</a> | <a href="https://whatpr.org/encoding/230/0ca7c4e...eade6ff.html" title="Last updated on Sep 21, 2020, 10:09 AM UTC (eade6ff)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -995,9 +995,16 @@ different format here, to be able to represent ranges.)
  <p>The <a>get an encoding</a> algorithm is to be used to turn a <a>label</a> into an
  <a for=/>encoding</a>.
 
- <p>Standards are to ensure that the I/O queues they pass to the <a for=/>encode</a> and
+ <p>Standards are to ensure that the input I/O queues they pass to the <a for=/>encode</a> and
  <a>UTF-8 encode</a> algorithms are effectively I/O queues of scalar values, i.e., they contain
  no <a>surrogates</a>.
+
+ <p>These hooks, with the exception of <a>BOM sniff</a>, will block until the input I/O queue has
+ been consumed in its entirety. In order to use the output tokens as they are pushed into the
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -995,9 +995,16 @@ different format here, to be able to represent ranges.)
...
  <a for=/>encoding</a>.
 
- <p>Standards are to ensure that the I/O queues they pass to the <a for=/>encode</a> and
+ <p>Standards are to ensure that the input I/O queues they pass to the <a for=/>encode</a> and
  <a>UTF-8 encode</a> algorithms are effectively I/O queues of scalar values, i.e., they contain
  no <a>surrogates</a>.
+
+ <p>These hooks, with the exception of <a>BOM sniff</a>, will block until the input I/O queue has
+ been consumed in its entirety. In order to use the output tokens as they are pushed into the
+ stream, callers are to invoke the hooks with an empty output I/O queue and read from it <a>in
```
annevk, line 1004: No newline inside `<a>`.

annevk (commented):
Thanks, looks good modulo nit. This isn't editorial per https://github.com/whatwg/meta/blob/master/COMMITTING.md. I do wonder if we should have something for these kind of changes as most folks would probably want to skip them too.

andreubotella:
In that case I've misunderstood what editorial means all along! 😅

Feel free to change the commit message when merging, then.

annevk (approved):
Thanks! I think I'd like to wait with landing until we fix building as it wouldn't end up synchronized with the server anyway.

PR #236, opened 2020-10-20
Editorial: make ISO-2022-JP encoder perform error state switch

### LINKED ISSUE #235: Encode hook not adequate
While working on https://github.com/whatwg/url/issues/557 I realized that the URL standard will have to invoke the Encoding Standard at a lower level of abstraction as it needs to deal with erroneous output (`&#...;`) differently from non-erroneous output. I.e., an error that results in `&#...;` might have to be percent-encoded, but a non-error `&#...;` sequence might not have to be.

So URL basically wants to invoke the encoder's handler directly I think. I don't really see a better abstraction as it needs to deal with errors in a very different way. I suppose we could make error handling a caller defined set of steps, but I don't really like the complexity of that.

`text/plain` form submission could in theory still use the current high-level encode hook, but I'm not sure it's worth saving just for that.

It also seems there's potentially quite a lot of other potential cleanup that would result from this (e.g., https://encoding.spec.whatwg.org/#concept-encoding-process no longer needs to handle "html").

Having "html" in the Encoding standard as well as this high-level hook was intentional as a way of limiting the amount of badness that could be introduced by consumers, but we will have to use review rather than abstractions for that instead (UTF-8 or die). It's unfortunate, but it might also make the Encoding standard a little leaner.

cc @andreubotella @hsivonen @achristensen07 @JKingweb

annevk:
This way you can invoke its handler directly without going the process algorithm as is needed to fix #235. This also avoids the need for "prepend" in the "process" algorithm.

Additionally, this commit adds two clarifying asserts to the "process" algorithm documenting what error modes can be in effect when.

***
<a href="https://whatpr.org/encoding/236.html" title="Last updated on Oct 20, 2020, 12:55 PM UTC (2c07038)">Preview</a> | <a href="https://whatpr.org/encoding/236/92569b5...2c07038.html" title="Last updated on Oct 20, 2020, 12:55 PM UTC (2c07038)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -304,6 +304,12 @@ happening.
  "<code>fatal</code>".
 
  <li><p>Assert: if <var>encoderDecoderInstance</var> is an <a for=/>encoder</a> instance,
+ <var>mode</var> is not "<code>replacement</code>".
+
+ <li><p>Assert: if <var>encoderDecoderInstance</var> is a <a for=/>decoder</a> instance,
+ <var>mode</var> is not "<code>html</code>".
+
+ <li><p>Assert: if <var>encoderDecoderInstance</var> is an <a for=/>encoder</a> instance,
  <var>item</var> is not a <a>surrogate</a>.
 
  <li><p>Let <var>result</var> be the result of running <var>encoderDecoderInstance</var>'s
@@ -340,9 +346,9 @@ happening.
    <dd><a>Push</a> U+FFFD to <var>output</var>.
 
    <dt>"<code>html</code>"
-   <dd><a>Prepend</a> U+0026, U+0023, followed by the
-   shortest sequence of <a>ASCII digits</a> representing <var>result</var>'s
-   <a>code point</a> in base ten, followed by U+003B to <var>input</var>.
+   <dd><a>Push</a> the result of <a>UTF-8 encoding</a> U+0026 (&amp;), U+0023 (#), followed by the
+   shortest sequence of <a>ASCII digits</a> representing <var>result</var>'s <a>code point</a> in
+   base ten, followed by U+003B (;) to <var>output</var>.
    <!-- &# ... ; -->
 
    <dt>"<code>fatal</code>"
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -340,9 +346,9 @@ happening.
    <dd><a>Push</a> U+FFFD to <var>output</var>.
 
    <dt>"<code>html</code>"
-   <dd><a>Prepend</a> U+0026, U+0023, followed by the
-   shortest sequence of <a>ASCII digits</a> representing <var>result</var>'s
-   <a>code point</a> in base ten, followed by U+003B to <var>input</var>.
+   <dd><a>Push</a> the result of <a>UTF-8 encoding</a> U+0026 (&amp;), U+0023 (#), followed by the
```
hsivonen, line 349: I suggest saying "0x26 (& encoded as UTF-8), 0x23 (# encoded as UTF-8)" and then "0x3B (; encoded as UTF-8)" to remove a layer of indirection.

hsivonen (changes requested):

encoding.bs
annevk: The problem is that we don't have a definition of https://infra.spec.whatwg.org/#ascii-digit that's byte-oriented. I could inline one, "the shortest sequence of 0x30 (0) to 0x39 (9), inclusive, representing ... in base ten?" if that's preferable. I guess this is also a good moment to start using https://infra.spec.whatwg.org/#code-point-value.

hsivonen: I'd prefer inlining such a definition over talking about UTF-8 encode here.

annevk: Gave that a go so you can see how it looks.

encoding.bs
```diff
@@ -2938,8 +2943,17 @@ and <var>code point</var>, runs these steps:
   <p class=note>If <var>pointer</var> is non-null, it is less than 8836 due to the nature of
   <a>index jis0208</a> and the <a>index pointer</a> operation.
 
- <li><p>If <var>pointer</var> is null, return <a>error</a> with
- <var>code point</var>.
+ <li>
+  <p>If <var>pointer</var> is null, then:
+
+  <ol>
+   <li><p>If <a>ISO-2022-JP encoder state</a> is <a lt="ISO-2022-JP encoder jis0208">jis0208</a>,
```
andreubotella, line 2950: Shouldn't this be

```suggestion
   <li><p>If <a>ISO-2022-JP encoder state</a> is not <a lt="ISO-2022-JP encoder ASCII">ASCII</a>,
```

annevk: We don't do a state switch earlier when it's an error and Roman and I don't see why we should here. I should probably write some tests to ensure implementations do the right thing though.

hsivonen: Please document (for non-browser users of the spec) that callers of the abstraction are responsible for emitting a replacement byte sequence that only uses the bytes that have the ASCII interpretation in the Roman state of ISO-2022-JP.

PR #237, opened 2020-10-21
Reorganize hooks for standards

annevk:
@andreubotella would appreciate your thoughts on this as well. I wanted to split this section into two given that I need two new algorithms for the URL standard (get an encoder and encode or fail) that both are somewhat wonky and probably need a bit of explanation and it seems bad to explain that together with all the UTF-8-centric primitives everyone ought to be using.

***
<a href="https://whatpr.org/encoding/237.html" title="Last updated on Oct 22, 2020, 9:15 AM UTC (789436e)">Preview</a> | <a href="https://whatpr.org/encoding/237/a5e3e11...789436e.html" title="Last updated on Oct 22, 2020, 9:15 AM UTC (789436e)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -263,11 +263,10 @@ happening.
 [[URL]]
 [[HTML]]
 
-<p>To <dfn id=concept-encoding-run for=encoding>run</dfn> an <a for=/>encoding</a>'s
-<a for=/>decoder</a> or <a for=/>encoder</a> <var>encoderDecoder</var> with input
-<a for=/>I/O queue</a> <var>input</var>, output
-<a for=/>I/O queue</a> <var>output</var>, and optional
-<a for=/>error mode</a> <var>mode</var>, run these steps:
+<p>To <dfn id=concept-encoding-run>run</dfn> an <a for=/>encoding</a>'s <a for=/>decoder</a> or
+<a for=/>encoder</a> <var>encoderDecoder</var> with input <a for=/>I/O queue</a> <var>input</var>,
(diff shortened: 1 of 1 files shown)


hsivonen (approved):
I take it that the new hook will be in a different PR.

encoding.bs
```diff
@@ -977,61 +973,28 @@ different format here, to be able to represent ranges.)
...
+ standards.
 
  <p>For decoding, <a>UTF-8 decode</a> is to be used by new formats. For identifiers or byte
  sequences within a format or protocol, use <a>UTF-8 decode without BOM</a> or
  <a>UTF-8 decode without BOM or fail</a>.
 
  <p>For encoding, <a>UTF-8 encode</a> is to be used.
 
- <p>Standards are strongly discouraged from using <a>decode</a>, <a for=/>encode</a>, and
- <a>BOM sniff</a>, except as needed for compatibility.
```
andreubotella, line 991: Maybe keep this paragraph in the note for legacy hooks.

andreubotella (commented):
LGTM

andreubotella:
Since #238 has "encode or fail" passing an encoder instance (rather than an encoder "class", so to speak) to "run", it's probably best to update that algorithm in this PR to take an instance and to change the hooks that call it to use "run a new instance of `encoding`'s decoder/encoder with..."

annevk:
Let's make those changes there if we keep that as initially proposed. I rather like putting the errors directly in the queue though.

PR #241, opened 2020-10-23
Clarify instance language around decoders and encoders

### LINKED ISSUE #240: Talk about instances of decoders and encoders
@andreubotella brought this up in #238 and it seems worth addressing, but not as part of that PR. (In particular I don't think it needs to impact a lot of the naming of terms, we just need to add instance in a few places.)

annevk:
And also stop defaulting error mode in "run" and "process".

Fixes #240.

r? @andreubotella

***
<a href="https://whatpr.org/encoding/241.html" title="Last updated on Oct 26, 2020, 4:38 PM UTC (abf398c)">Preview</a> | <a href="https://whatpr.org/encoding/241/c55584b...abf398c.html" title="Last updated on Oct 26, 2020, 4:38 PM UTC (abf398c)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -1569,6 +1569,8 @@ method steps are:
  <a lt="get a reference to the buffer source">getting a reference to the bytes held by</a>
  <var>destination</var>.
 
+ <li><p>Let <var>encoder</var> be an instance of the <a>UTF-8 encoder</a>.
+
  <li>
   <p>Let <var>unused</var> be the <a for=/>I/O queue</a> of scalar values « <a>end-of-queue</a> ».
 
@@ -1584,8 +1586,8 @@ method steps are:
   <ol>
    <li><p>Let <var>item</var> be the result of <a>reading</a> from <var>source</var>.
 
-   <li><p>Let <var>result</var> be the result of running the <a>UTF-8 encoder</a>'s <a>handler</a>
-   on <var>unused</var> and <var>item</var>.
+   <li><p>Let <var>result</var> be the result of running <var>encoder</var>'s <a>handler</a> on
+   <var>unused</var> and <var>item</var>.
 
    <li><p>If <var>result</var> is <a>finished</a>, then <a for=iteration>break</a>.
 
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -1597,8 +1586,8 @@ method steps are:
   <ol>
    <li><p>Let <var>item</var> be the result of <a>reading</a> from <var>source</var>.
 
-   <li><p>Let <var>result</var> be the result of running the <a>UTF-8 encoder</a>'s <a>handler</a>
-   on <var>unused</var> and <var>item</var>.
+   <li><p>Let <var>result</var> be the result of running <var>encoder</var>'s <a>handler</a> on
```
andreubotella, line 1589:
```suggestion
   <li><p>Let <var>result</var> be the result of <a>running</a> <var>encoder</var>'s <a>handler</a> on
```

andreubotella (commented):
You're missing "an instance of" in UTF-8 decode without BOM or fail. Also in TextDecoderCommon's "decoder" and TextEncoderStream's "encoder" associated fields.

I'd also add some clarification on the fact that encoders and decoders are abstract and their state goes on instances:

> Each encoding has an associated decoder and most of them have an associated encoder. Encoders and decoders are stateless, but they can be used to create instances which might have state, and which have a handler algorithm.

encoding.bs
annevk: This does not want to invoke "run". It's the one case where neither "run" nor "process" is used. It's using run in the normal sense of the word. (Perhaps we should rename "run"...)

annevk:
I guess the one thing this does not address is that where we define decoders and encoders we do not spell out "instance" at the moment. We could, but I'd like to tackle that separately I think.

andreubotella (approved):
The note to "encode or fail" has "keep an encoder alive" and "supplying the same encoder".

The domintros also have language like "running encoding's decoder", but since "running" isn't linked to the algorithm (see #245) and encoder/decoder instances can be seen as equivalent to a generator, I guess that's fine.

Other than that, LGTM.

PR #244, opened 2020-10-26
Add some minor clarifications.

### LINKED ISSUE #239: Prepending or conversion of a sequence possibly not exact enough explained
In section 3 'Terminology' the idea of prepending a sequence of items to an I/O queue, or converting a list of items into an I/O queue is explained.  The described mechanisms work only, if the particular end-of-queue item is not present in the source.  Even if we accept for a moment, that an unreliable external source can never accidentally add an end-of-queue item to a sequence or a list, a sequence or a list may still contain this item.  This is because an I/O queue itself is a list, which in turn is an ordered sequence per definition.  So speaking of a sequence/list of items does not preclude the presence of an end-of-queue.
In
> To prepend a sequence of items, insert those items, in the given order, before the first item in the queue.

the possibly present end-of-queue item MUST NOT be inserted into a destination I/O queue, unless it is empty.
Similar for conversion.  IMO one should adapt the wording here to account for this exceptional situation.

Wolf Lammen

### LINKED ISSUE #243: Questionable conversion in BOM Sniff
In section 6.1 under the topic **BOM sniff** I read

> 1. Let BOM be the result of peeking 3 bytes from ioQueue, converted to a byte sequence.

The conversion makes no sense here.  The result of a **peek** is a **list** without **end-of-queue**.  In particular, it is not an **I/O queue** for that reason.  Now **conversions** are only declared if an **I/O queue** is involved, either as a source a destination.  This is for one formally not the case here, since the source is a **list** of bytes and the destination a sequence of bytes. And second a **list** of bytes is already a sequence of bytes, so semantically no conversion is needed at all.

Wolf Lammen

andreubotella:
Closes #239, closes #243.

cc @wlammen

***
<a href="https://whatpr.org/encoding/244.html" title="Last updated on Oct 26, 2020, 11:06 AM UTC (638d76a)">Preview</a> | <a href="https://whatpr.org/encoding/244/c55584b...638d76a.html" title="Last updated on Oct 26, 2020, 11:06 AM UTC (638d76a)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -199,11 +199,11 @@ an I/O queue "<code>&amp;#128169; hello world</code>". The next item to be read 
 <code>&amp;</code>. <!-- 💩 -->
 
 <p>To <dfn for="from I/O queue">convert</dfn> an <a for=/>I/O queue</a> <var>ioQueue</var> into a
-<a for=/>list</a>, <a>string</a> or <a>byte sequence</a>, return the result of
+<a for=/>list</a>, <a>string</a>, or <a>byte sequence</a>, return the result of
 <a for="I/O queue">reading</a> an indefinite number of <a for=list>items</a> from
 <var>ioQueue</var>.
 
-<p>To <dfn for="to I/O queue">convert</dfn> a <a for=/>list</a>, <a>string</a> or
+<p>To <dfn for="to I/O queue">convert</dfn> a <a for=/>list</a>, <a>string</a>, or
 <a>byte sequence</a> <var>input</var> into an <a for=/>I/O queue</a>, run these steps:
 
 <ol>
--- a/encoding.bs
+++ b/encoding.bs
@@ -207,7 +207,8 @@ an I/O queue "<code>&amp;#128169; hello world</code>". The next item to be read 
 <a>byte sequence</a> <var>input</var> into an <a for=/>I/O queue</a>, run these steps:
 
 <ol>
- <li><p>Assert: <var>input</var> does not <a for=list>contain</a> <a>end-of-queue</a>.
+ <li><p>Assert: if <var>input</var> is a <a for=/>list</a>, then it does not <a for=list>contain</a>
+ <a>end-of-queue</a>.
 
  <li><p>Return an <a for=/>I/O queue</a> containing the <a for=list>items</a> in <var>input</var>,
  in order, followed by <a>end-of-queue</a>.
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -201,10 +201,17 @@ an I/O queue "<code>&amp;#128169; hello world</code>". The next item to be read
...
-<a>byte sequence</a> <var>input</var> into an <a for=/>I/O queue</a>, return an
-<a for=/>I/O queue</a> containing the <a for=list>items</a> in <var>input</var>, in order, followed
-by <a>end-of-queue</a>.
+<var>ioQueue</var>.
+
+<p>To <dfn for="to I/O queue">convert</dfn> a <a for=/>list</a>, <a>string</a> or
+<a>byte sequence</a> <var>input</var> into an <a for=/>I/O queue</a>, run these steps:
+
+<ol>
+ <li><p>Assert: <var>input</var> does not <a for=list>contain</a> <a>end-of-queue</a>.
```
annevk, line 210: This needs a conditional for input being a list. Or we do it similar to prepend and say "a list not containing end-of-queue" in the algorithm "header".

encoding.bs
```diff
@@ -999,11 +1006,11 @@ different format here, to be able to represent ranges.)
...
 
 <ol>
- <li><p>Let <var>buffer</var> be the result of <a for="I/O queue">reading</a> three bytes from
- <var>ioQueue</var>.
+ <li><p>Let <var>buffer</var> be the result of <a for="I/O queue">peeking</a> three bytes from
+ <var>ioQueue</var>, converted to a byte sequence.
 
- <li><p>If <var>buffer</var> does not match 0xEF 0xBB 0xBF, <a>prepend</a> <var>buffer</var> to
- <var>ioQueue</var>.
+ <li><p>If <var>buffer</var> matches 0xEF 0xBB 0xBF, <a for="I/O queue">read</a> three bytes from
```
annevk, line 1012: Slight preference for using "is" here over "matches". As well ", then ...".

annevk (commented):
Thanks Andreu for the quick fix!

encoding.bs
andreubotella: I tried having the assertion in the algorithm header, and it did not flow well ("to convert one of a list not containing end-of-string, a string or a byte sequence `input`"). But sure, I'll add the conditional.

encoding.bs
```diff
@@ -201,10 +201,17 @@ an I/O queue "<code>&amp;#128169; hello world</code>". The next item to be read
 <p>To <dfn for="from I/O queue">convert</dfn> an <a for=/>I/O queue</a> <var>ioQueue</var> into a
 <a for=/>list</a>, <a>string</a> or <a>byte sequence</a>, return the result of
 <a for="I/O queue">reading</a> an indefinite number of <a for=list>items</a> from
-<var>ioQueue</var>. To <dfn for="to I/O queue">convert</dfn> a <a for=/>list</a>, <a>string</a> or
-<a>byte sequence</a> <var>input</var> into an <a for=/>I/O queue</a>, return an
-<a for=/>I/O queue</a> containing the <a for=list>items</a> in <var>input</var>, in order, followed
-by <a>end-of-queue</a>.
+<var>ioQueue</var>.
+
+<p>To <dfn for="to I/O queue">convert</dfn> a <a for=/>list</a>, <a>string</a> or
```
annevk, line 206: Please add an Oxford comma before "or".

annevk (approved):
Will give Wolf some time to take a look as well.

wlammen:
I don't have any objections.  Thanks for inviting me for a review.

Issue #171, opened 2019-01-23
Big5 encoding mishandles some trailing bytes, with possible XSS


mihnita:
There are some sequences of bytes that are valid lead-trailing according to the description at https://encoding.spec.whatwg.org/#big5-decoder, but don't have a corresponding Unicode codepoint in the `index-big5.txt` mapping table.

This in this case the first byte is converted to U+FFFD, but the second one is left "as is". In some cases that second byte can be backslash (`5C`), which can be used to escape end of strings in JavaScript and potentially resulting in XSS.

Example (attached): the `83 5C` sequence
According to the algorithm at https://encoding.spec.whatwg.org/#big5-decoder
* the first byte is a lead byte (case 5, byte is in the range 0x81 to 0xFE, inclusive)
* for the second byte we have case 3
    * 3.1. Let offset be 0x40 (byte is less than 0x7F)
    * 3.2. byte is in the range 0x40 to 0x7E, inclusive => set pointer to `(lead − 0x81) × 157 + (byte − offset)`. The result of that is `(0x83 - 0x81) * 157 + (0x5C - 0x40)` which is `0x156`
   * there is no mapping for 0x156 in `index-big5.txt`, and because the `byte is an ASCII byte` we `prepend byte to stream` (case 3.6) and `Return error` (case 3.7)

The end result is a FFFD (from the error) followed by a 5C (the trailing byte, "as is")

You can see this in the attached file.
When opened in both Chrome and Firefox the text is rendered as the "Unicode REPLACEMENT CHARACTER" (correct) followed by a back-slash (incorrect).

This is a valid lead-trail byte sequence that should either be replaced by one single U+FFFD character, or by two U+FFFD characters. But the definitely the trailing byte should not be left "as is"

The possible exploit can use the trailing byte (which is backslash) to escape the end of a string, for example.
Checking the console of Firefox you will see the 'SyntaxError: "" string literal contains an unescaped line break' message. In Chrome the message is 'Uncaught SyntaxError: Invalid or unexpected token'

I did not check, but this might also happen in other DBCS (Double Byte Characters Sets) that have the second byte in the ASCII range (for instance in Shift JIS?).

[big5.zip](https://github.com/whatwg/encoding/files/2788604/big5.zip)

mihnita:
Thank you for the quick answer.

**==== TLDR ====**

`82 22` which is an invalid Shift JIS byte sequence.
`83 5C` is a valid Big 5 sequence, with not Unicode mapping.

[...]

PR #254, opened 2021-02-26
Document gb18030 decoder end-of-queue oddity

### LINKED ISSUE #253: End-of-queue during decoding of GB18030 should not mask ASCII characters.
https://encoding.spec.whatwg.org/commit-snapshots/4d54adce6a871cb03af3a919cbf644a43c22301a/#gb18030-decoder

> If byte is end\-of\-queue, and gb18030 first, gb18030 second, or gb18030 third is not 0x00, set gb18030 first, gb18030 second, and gb18030 third to 0x00, and return error\. 

I think this violates the requirements in the [Security Background section](https://encoding.spec.whatwg.org/#security-background)
> Decoders of encodings that use multiple bytes for scalar values now require that in case of an illegal byte combination, a scalar value in the range U+0000 to U+007F, inclusive, cannot be “masked”.

In particular, the input sequence 0x81 0x30 should, by my reading of the sentence quoted above, produce U+FFFD U+0030 but according to the specification, only a single U+FFFD is produced.

The Rust crate `encoding_rs` agrees with the specification.
```rust
use encoding_rs::*;

fn main() {
    let (output, replacements) = GB18030.decode_without_bom_handling(&[0x81, 0x30]);
    assert!(replacements);
    assert_eq!(output, "\u{FFFD}0");
}
```
The second assertion fails because the output contains just the U+FFFD.

Chrome, Firefox, and Safari all agree with `encoding_rs`. E.g., appending the byte sequence 0x81 0x30 to
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset=gb18030>
  </head>
  <body>
    <span id=bug>
```
results in the `span` element containing just U+FFFD.

Maybe this masking of an ASCII character at the end of the input is fine and the security background should be updated instead to note that fact.

A similar issue arises with the byte sequence 0x81 0x30 0x81 which I'd expect to be U+FFFD U+0030 U+FFFD but instead decodes to a single replacement U+FFFD.

annevk:
Closes #253.

***
<a href="https://whatpr.org/encoding/254.html" title="Last updated on Feb 26, 2021, 11:19 AM UTC (394c83a)">Preview</a> | <a href="https://whatpr.org/encoding/254/4d54adc...394c83a.html" title="Last updated on Feb 26, 2021, 11:19 AM UTC (394c83a)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -42,34 +42,34 @@ specification does not provide a mechanism for extending any aspect of encodings
 
 <h2 id=security-background>Security background</h2>
 
-<p>There is a set of encoding security issues when the producer and consumer do not agree
-on the encoding in use, or on the way a given encoding is to be implemented. For instance,
-an attack was reported in 2011 where a <a>Shift_JIS</a> lead byte 0x82 was used to
-“mask” a 0x22 trail byte in a JSON resource of which an attacker could control some field.
-The producer did not see the problem even though this is an illegal byte combination. The
-consumer decoded it as a single U+FFFD and therefore changed the overall interpretation as
-U+0022 is an important delimiter. Decoders of encodings that use multiple bytes for scalar
-values now require that in case of an illegal byte combination, a scalar value in the
-range U+0000 to U+007F, inclusive, cannot be “masked”. For the aforementioned sequence the
-output would be U+FFFD U+0022.
+<p>There is a set of encoding security issues when the producer and consumer do not agree on the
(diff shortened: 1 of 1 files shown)


ricea (approved):
lgtm

hsivonen:
LGTM as a gb18030-specific documentation step, but we still haven't solved the issue opposite to what's being documented here: #171.

PR #260, opened 2021-05-28
Editorial: Use "set up" for TransformStreams rather than "create"

andreubotella:
Follows whatwg/streams#1110.

I just noticed that the Encoding standard was failing to build because of these changes to the Streams standards.

***
<a href="https://whatpr.org/encoding/260.html" title="Last updated on Jun 21, 2021, 8:56 AM UTC (cb41266)">Preview</a> | <a href="https://whatpr.org/encoding/260/75b988c...cb41266.html" title="Last updated on Jun 21, 2021, 8:56 AM UTC (cb41266)">Diff</a>

### DIFF
--- a/.htmlcheckerfilter
+++ b/.htmlcheckerfilter
@@ -1,3 +1,4 @@
 .*Text run is not in Unicode Normalization Form C.*
 .*Document uses the Unicode Private Use Area.*
 .*File was not checked. Files must have .html, .xhtml, .htm, or .xht extensions.*
+.*Possible misuse of “aria-label”. (If you disagree with this warning, file an issue report or send e-mail to www-validator@w3.org.).*
--- a/.htmlcheckerfilter
+++ b/.htmlcheckerfilter
@@ -1,4 +1,4 @@
 .*Text run is not in Unicode Normalization Form C.*
 .*Document uses the Unicode Private Use Area.*
 .*File was not checked. Files must have .html, .xhtml, .htm, or .xht extensions.*
-.*Possible misuse of “aria-label”. (If you disagree with this warning, file an issue report or send e-mail to www-validator@w3.org.).*
+.*Possible misuse of “aria-label”. \(If you disagree with this warning, file an issue report or send e-mail to www-validator@w3.org.\).*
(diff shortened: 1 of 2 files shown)


andreubotella:
The build failures are caused by the visualization and BMP coverage tables (https://encoding.spec.whatwg.org/windows-1252.html for example), [since they have `aria-label` attributes on `<td>` elements](https://github.com/whatwg/encoding/blob/main/visualize.py#L208), which as of validator/validator@726e7593b1eda345392a080ba179abbc2c6352ca causes the HTML checker to fail.

@sideshowbarker Changing the `aria-label` attributes to `title` would make the checker work, but I'm not sure whether that'd be fine accessibility-wise.

cc @annevk

domenic (approved):
Thanks for catching this!

annevk:
Thanks! I suggest we sort #261 first. I also think that we should land this prefixed with "Editorial:" as it ends up changing the prose, not just the references.

sideshowbarker:
> @sideshowbarker Changing the `aria-label` attributes to `title` would make the checker work, but I'm not sure whether that'd be fine accessibility-wise.

Per a strict reading of the current wording of the implementation requirements at https://w3c.github.io/html-aam/#other-tabular-data-elements, I believe it would not be wrong accessibility-wise to change the `aria-label` attributes to `title`. However, it seems like that current wording in the AAM spec may not actually makes much sense and may instead be a spec bug. I mean, why would it make sense to make use of any `title` attribute on the `td`/`th`/`tr`, but not make use of any `aria-lable` attribute?

So I raised https://github.com/w3c/html-aam/issues/335 to get clarification.

If a change does end up getting made to the AAM spec, then I guess it would be appropriate for me to change the checker back to allowing `aria-label` on `td`/`th`/`tr` without warning. I say *“I guess”* because https://html5accessibility.com/stuff/2020/11/07/not-so-short-note-on-aria-label-usage-big-table-edition/ says that `td`/`th`/`tr` for “aria-label use not well supported” in UAs/AT. So I take that to mean those `aria-label` attributes would in practice not actually get exposed as accessible names in practice in current UAs/AT. And I suspect that `title` wouldn’t either. So I really don’t know what’s best to do here…

PR #284, opened 2022-04-22
Static methods for TextEncoder and TextDecoder

### LINKED ISSUE #267: Add a static decode and encode method to `TextEncoder` and `TextDecoder`
The large majority of users of `TextEncoder` and `TextDecoder` do not make use of streaming capabilities. Instead they just need to decode a single chunk, or encode a single chunk. We should consider adding a static utility method to `TextEncoder` / `TextDecoder` to cater to this use case.

```js
// current
const data = new TextDecoder().decode(new Uint8Array(/** data here */));

// proposed
const data = TextDecoder.decode(new Uint8Array(/** data here */));
```

I propose the following API be added:

```webidl
dictionary TextDecoderOptionsWithLabel : TextDecoderOptions {
  DOMString label = "utf-8";
};

interface TextDecoder {
  static USVString decode(optional [AllowShared] BufferSource input, optional TextDecoderOptionsWithLabel options = {});
}

interface TextEncoder {
  static Uint8Array encode(optional USVString input = "");
}

```

The exact name of the method should be something different though - the static method and prototype method would name clash in this example which is confusing.

Internally this would behave exactly like a single use non streaming encoder.

lucacasonato:
NOTE: this PR is currently blocked on a technicality (https://github.com/tabatkins/bikeshed/issues/2270).

This PR adds support for static `TextDecoder.decode`, `TextEncoder.encode` and
`TextEncoder.encodeInto` methods. These do not add new functionality, rather
just acting as useful helpers for "one time" text encoding/decoding operations.

Closes #267.

  - Deno
  - …
- [ ] [Tests](https://github.com/web-platform-tests/wpt) are written and can be
      reviewed and commented upon at:
  - …
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests)
      are filed:
  - Chrome: …
  - Firefox: …
  - Safari: …
  - Deno: …
  - Node.js: …


encoding.bs
```diff
@@ -1333,6 +1333,10 @@ dictionary TextDecoderOptions {
   boolean ignoreBOM = false;
 };
 
+dictionary TextDecoderOptionsWithLabel : TextDecoderOptions {
+  DOMString label = "utf-8";
```
annevk, line 1337: I wonder if we can get away with just supporting UTF-8 here. If you need more, you get the streaming API. Perhaps not even support `fatal` or `ignoreBOM`. Also makes it easier to implement without a lot of IDL bridging.

encoding.bs
```diff
@@ -1351,13 +1357,17 @@ TextDecoder includes TextDecoderCommon;
 initially false.
 
 <dl class=domintro>
+ <dt>{{TextDecoder}} . <a method for=TextDecoder lt=decode()>decode([<var>input</var> [, <var>options</var>]])</a></code>
```
annevk, line 1360:
```suggestion
 <dt><code><a>TextDecoder</a> . <a method for=TextDecoder lt=decode()>decode([<var>input</var> [, <var>options</var>]])</a></code>
```

encoding.bs
```diff
@@ -1351,13 +1357,17 @@ TextDecoder includes TextDecoderCommon;
...
+ <dd>
+  <p>Returns the result of running <a for=TextDecoderCommon>encoding</a>'s <a for=/>decoder</a>.
```
annevk, line 1362: Below this should probably be prefixed with _decoder_'s and here it needs some rewording as there is no instance to speak of.

encoding.bs
```diff
@@ -1528,11 +1588,7 @@ requires buffering of scalar values.
  <var>destination</var>.
 </dl>
 
-<p>The
-<dfn constructor for=TextEncoder lt=TextEncoder() id=dom-textencoder><code>new TextEncoder()</code></dfn>
-constructor steps are to do nothing.
-
-<p>The <dfn method for=TextEncoder><code>encode(<var>input</var>)</code></dfn> method steps are:
+<p>To <a>encode an input</a> given a <a>USVString</a> <var>input</var>, run the following steps:
```
annevk, line 1591: Needs `<dfn>`.

encoding.bs
```diff
@@ -1563,9 +1619,8 @@ constructor steps are to do nothing.
   </ol>
 </ol>
 
-<p>The
-<dfn method for=TextEncoder><code>encodeInto(<var>source</var>, <var>destination</var>)</code></dfn>
-method steps are:
+<p>To <a>encode a source into a destination</a> given a <a>USVString</a> <var>source</var>, and a
```
annevk, line 1622: Needs `<dfn>`.

jasnell:
@lucacasonato ... are you still wanting to push this forward. I'm interested in implementing this for both workers and node.js

PR #309, opened 2023-04-05
Editorial: use leading and trailing surrogate

annevk:
Also introduce an operation to obtain a scalar value from surrogates.

Eventually the lead/trail byte stuff needs to be made consistent with this as well.

***
<a href="https://whatpr.org/encoding/309.html" title="Last updated on Oct 4, 2024, 9:30 AM UTC (8a5f3b0)">Preview</a> | <a href="https://whatpr.org/encoding/309/2c3853e...8a5f3b0.html" title="Last updated on Oct 4, 2024, 9:30 AM UTC (8a5f3b0)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -1855,8 +1855,8 @@ TextEncoderStream includes GenericTransformStream;
  <dt><dfn for=TextEncoderStream>encoder</dfn>
  <dd>An <a for=/>encoder</a> instance.
 
- <dt><dfn for=TextEncoderStream>pending high surrogate</dfn>
- <dd>Null or a <a for=/>surrogate</a>, initially null.
+ <dt><dfn for=TextEncoderStream id=textencoderstream-pending-high-surrogate>leading surrogate</dfn>
+ <dd>Null or a <a for=/>leading surrogate</a>, initially null.
 </dl>
 
 <p class="note no-backref">A {{TextEncoderStream}} object offers no <var>label</var> argument as it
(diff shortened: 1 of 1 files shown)


encoding.bs
```diff
@@ -1975,26 +1975,26 @@ constructor steps are:
...
+   0x10000 + ((<var>leadingSurrogate</var> &minus; 0xD800) &lt;&lt; 10) + (<var>item</var> &minus; 0xDC00).
 
    <li><p><a>Prepend</a> <var>item</var> to <var>input</var>.
 
    <li><p>Return U+FFFD.
   </ol>
 
- <li><p>If <var>item</var> is in the range U+D800 to U+DBFF, inclusive, then set <a>pending high
- surrogate</a> to <var>item</var> and return <a>continue</a>.
+ <li><p>If <var>item</var> is a <a>leading surrogate</a>, then set <a>leading surrogate</a> to
```
ricea, line 1994: I think both these `<a>leading surrogate</a>` will link to the same place, which seems wrong.

ricea (commented):
Only the title says "pending surrogate"; should that be "trailing surrogate" instead?

encoding.bs
annevk: Should be fixed once https://github.com/whatwg/infra/commit/0dba3a29da6e071a30e06bb837e282eca33efbc9 is indexed, but I should probably use explicit `for` as well.

ricea (approved):
lgtm

PR #326, opened 2024-03-03
Clarify insertion order when prepending multiple bytes at once to ioQueue

### LINKED ISSUE #325: Ambiguous wording in GB18030 decoder

I am struggling to understand what "Prepend *gb18030 second* followed by *byte* to ioQueue" exactly means in [gb18030 decoder](https://encoding.spec.whatwg.org/#gb18030-decoder) step 4.2.

Does it mean to prepend a sequence containing *gb18030 second* and *byte* or does it mean first prepend *gb18030 second*, then prepend *byte*? This is the only place where this wording is used in the context of "prepend". All the other usages of "prepend" are unambiguous.

As a side note, I propose to entirely get rid of the concept of prepending a sequence. It is used only in one place, in step 5.5 of [utf-16-decoder](https://encoding.spec.whatwg.org/#shared-utf-16-decoder), and it only clutters understanding. It would be much clearer to just say "if *UTF-16BE decoder* is true prepend *byte2*, then prepend *byte1*, otherwise, prepend *byte1*, then prepend *byte2*". The note just above [4. Encodings](https://encoding.spec.whatwg.org/#encodings) says that "prepend operation is an internal detail of the algorithms in this specification, and is not to be used by other standards", so it looks safe to remove this concept.

stasoid:
Main source of confusion for me was the absence of the keyword "sequence". This PR fixes that.
Fixes #325.

***
<a href="https://whatpr.org/encoding/326.html" title="Last updated on Mar 3, 2024, 9:59 AM UTC (0575ad0)">Preview</a> | <a href="https://whatpr.org/encoding/326/b53fc98...0575ad0.html" title="Last updated on Mar 3, 2024, 9:59 AM UTC (0575ad0)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -2339,7 +2339,7 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
     <p>If <var>byte</var> is not in the range 0x30 to 0x39, inclusive, then:
 
     <ol>
-     <li><p><a>Prepend</a> <a>gb18030 second</a>, <a>gb18030 third</a>, and <var>byte</var> to
+     <li><p><a>Prepend</a> a sequence of <a>gb18030 second</a>, <a>gb18030 third</a>, and <var>byte</var> to
      <var>ioQueue</var>.
 
      <li><p>Set <a>gb18030 first</a>, <a>gb18030 second</a>, and <a>gb18030 third</a> to 0x00.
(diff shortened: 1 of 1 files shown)


stasoid:
I [signed up](https://github.com/whatwg/participant-data/commit/428a8a16e3ac318c7cdd671acd8c60d31fb71a60), do I have to do something else to unblock the PR?

stasoid:
main branch is broken: `LINK ERROR: Ambiguous for-less link for 'name'`

stasoid:
Guys, your main branch is broken - `make` fails with `LINK ERROR: Ambiguous for-less link for 'name'`. That's why my PR also fails - I didn't touch any links, changes only in plain text.

annevk:
I went for something slightly different as sequence itself is not really an established term outside of IDL: #328. Let me know what you think.

stasoid:
Yes, this edit clarifies the order, thank you.

annevk:
Thanks! Good call on the example. Pushed a fix for that too.

PR #328, opened 2024-03-11
Editorial: make prepend more clearly operate on a list

### LINKED ISSUE #325: Ambiguous wording in GB18030 decoder

annevk:
Fixes #325 and closes #326.

***
<a href="https://whatpr.org/encoding/328.html" title="Last updated on Mar 13, 2024, 8:52 AM UTC (49ea500)">Preview</a> | <a href="https://whatpr.org/encoding/328/b53fc98...49ea500.html" title="Last updated on Mar 13, 2024, 8:52 AM UTC (49ea500)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -190,7 +190,7 @@ from an <a for=/>I/O queue</a> <var>ioQueue</var>, run these steps:
 
 <p>To <dfn id=concept-stream-prepend for="I/O queue">prepend</dfn> an <a for=list>item</a> other
 than <a>end-of-queue</a> to an <a for=/>I/O queue</a>, perform the normal <a for=/>list</a>
-<a for=list>prepend</a> operation. To prepend a sequence of items not containing
+<a for=list>prepend</a> operation. To prepend a <a for=/>list</a> of <a for=list>items</a> excluding
 <a>end-of-queue</a>, insert those items, in the given order, before the first item in the queue.
 
 <p class=example id=example-tokens>Inserting the sequence of scalar value items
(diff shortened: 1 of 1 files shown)


hsivonen:
r+, but it would be worth considering if it should be called "restore" or "unread" to make the point that "prepend" is only ever used to put bytes back as if they hadn't been read and is never about any kind of other operation.

hsivonen (approved):

annevk:
Good call. I've pushed a fix for that as well as for the feedback in #326 about the example.

stasoid:
Cool, that's perfect.

Issue #312, opened 2023-06-06
Reflect changes in GB 18030-2022


achristensen07:
https://encoding.spec.whatwg.org/index-gb18030.txt contains 18 code points that have been changed by GB 18030-2022.  We should probably update.

domenic:
Related: #27 #57. It's not clear whether web browsers want to update these, or leave legacy encodings forever-unchanged.

litherum:
We, at least, definitely want to update. The source of truth for this encoding has updated it. There needs to be a consistent behavior between the web browser on a platform and native apps on the platform. As the adage goes, "the future is longer than the past" and there will be more content produced with the new mappings than there is existing content.

hsivonen:
@litherum Did WebKit implement [Unicode Technical Committee recommendation on this topic](https://www.unicode.org/L2/L2023/23003r-gb18030-recommendations.pdf)?

litherum:
Yes.

achristensen07:
We ended up moving away from that recommendation and going with the exact GB 18030-2022 mappings instead.

achristensen07:
We have since moved to the Unicode recommendation.  Either way, something related to GB 18030-2022 should probably be reflected in this standard.

annevk:
These are the Unicode recommendations:
```
# GB 18030 & Unicode Standard Transcoding Recommendations
#
# Transcoding Between GB 18030 & Unicode Standard
0xA6D9 <-> U+FE10
0xA6DA <-> U+FE12
0xA6DB <-> U+FE11
0xA6DC <-> U+FE13
0xA6DD <-> U+FE14
0xA6DE <-> U+FE15
0xA6DF <-> U+FE16
0xA6EC <-> U+FE17
0xA6ED <-> U+FE18
0xA6F3 <-> U+FE19
0xFE59 <-> U+9FB4
0xFE61 <-> U+9FB5
0xFE66 <-> U+9FB6
0xFE67 <-> U+9FB7
0xFE6D <-> U+9FB8
0xFE7E <-> U+9FB9
0xFE90 <-> U+9FBA
0xFEA0 <-> U+9FBB
#
# Transcoding From Unicode Standard to GB 18030
U+E78D -> 0xA6D9
U+E78E -> 0xA6DA
U+E78F -> 0xA6DB
U+E790 -> 0xA6DC
U+E791 -> 0xA6DD
U+E792 -> 0xA6DE
U+E793 -> 0xA6DF
[...]

PR #335, opened 2024-09-17
Support GB18030-2022

annevk:
One legacy encoding was updated and relevant regulation requires software to match. As such the Encoding Standard should match as well. This aims to make the minimum number of changes necessary and does not impact GBK, only gb18030.

Updated tests are upstreamed in https://github.com/web-platform-tests/wpt/pull/48221. (See also the comment below.)

   * WebKit
   * …
   * See above.
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: …
   * Gecko: …
   * WebKit: N/A
   * Deno: …
   * Node.js: …
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***
<a href="https://whatpr.org/encoding/335.html" title="Last updated on Sep 17, 2024, 12:08 PM UTC (e43cf40)">Preview</a> | <a href="https://whatpr.org/encoding/335/e20f586...e43cf40.html" title="Last updated on Sep 17, 2024, 12:08 PM UTC (e43cf40)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -835,7 +835,8 @@ specification, excluding <a>index single-byte</a>, which have their own table:
   <td>This matches the GB18030-2005 standard for code points encoded as two bytes, except for
   0xA3 0xA0 which maps to U+3000 to be compatible with deployed content. This index covers the
   CJK Unified Ideographs block of Unicode in its entirety. Entries from that block that are above or
-  to the left of (the first) U+3000 in the visualization are in the Unicode order.
+  to the left of (the first) U+3000 in the visualization are in the Unicode order. (Support for the
+  GB18030-2022 standard is handled separately to avoid impacting <a>GBK</a>.)
   <!-- https://bugzilla.mozilla.org/show_bug.cgi?id=131837
        https://bugs.webkit.org/show_bug.cgi?id=17014
        https://www.w3.org/Bugs/Public/show_bug.cgi?id=25396
@@ -2298,13 +2299,14 @@ historically this might have been the case for <a>ISO-8859-6</a> and
 
 <h4 id=gbk-decoder dfn export>GBK decoder</h4>
 
-<p><a>GBK</a>'s <a for=/>decoder</a> is <a>gb18030</a>'s <a for=/>decoder</a>.
+<p><a>GBK</a>'s <a for=/>decoder</a> is <a>gb18030</a>'s <a for=/>decoder</a> with its
+<a for="gb18030 decoder">is GBK</a> set to true.
 
 
 <h4 id=gbk-encoder dfn export>GBK encoder</h4>
 
-<p><a>GBK</a>'s <a for=/>encoder</a> is <a>gb18030</a>'s <a for=/>encoder</a>
-with its <a>is GBK</a> set to true.
+<p><a>GBK</a>'s <a for=/>encoder</a> is <a>gb18030</a>'s <a for=/>encoder</a> with its
+<a for="gb18030 encoder">is GBK</a> set to true.
 
 <p class="note no-backref">Not fully aliasing <a>GBK</a> with <a>gb18030</a>
 is a conservative move to decrease the chances of breaking legacy servers and other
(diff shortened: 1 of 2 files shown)


annevk:
The tests are upstreamed in https://github.com/web-platform-tests/wpt/pull/48221 and I noticed there that Chromium previously imported WebKit's changes and exported them to web-platform-tests. This was done by @yoshisatoyanagisawa and @tkent-google: https://github.com/web-platform-tests/wpt/commit/2ddb8af864f5935c25d84a3d7b23f9fbb0af087d.

Thus it seems Chromium has partial compliance with these changes. But not the most recent changes where WebKit (once again) aligned with the Unicode recommendation around gb18030.

One thing that might be good to test is that GBK is not impacted by these new mappings. (At least, I've been assuming we don't want that.)

@achristensen07 @ricea @hsivonen @domenic thoughts and review appreciated!

encoding.bs
```diff
@@ -2347,14 +2350,162 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
...
     </ol>
 
-   <li><p>Let <var>code point</var> be the <a>index gb18030 ranges code point</a> for
+   <li><p>Let <var>code point</var> be null.
+
+   <li>
+    <p>If <a for="gb18030 decoder">is GBK</a> is false and there is a row in the table below whose
+    first column is <a>gb18030 first</a>, second column is <a>gb18030 second</a>, third column is
+    <a>gb18030 third</a>, and fourth column is <var>byte</var>, then set <var>code point</var> to
+    the fifth column on the same row:
```
annevk, line 2359: Based on test results in Firefox I calculated the first override here and it appears that this is already the existing behavior of gb18030 as written in the standard. As such this table isn't needed. It's not clear to me why the Unicode recommendation calls out these four-byte sequences mapping to code points if that matches existing behavior.

encoding.bs
```diff
@@ -2431,8 +2688,131 @@ consumers of content generated with <a>GBK</a>'s <a for=/>encoder</a>.
...
 
- <li><p>If <a>is GBK</a> is true and <var>code point</var> is
- U+20AC, return byte 0x80.
+ <li><p>If <a for="gb18030 encoder">is GBK</a> is true and <var>code point</var> is U+20AC, then
+ return byte 0x80.
+
+ <li>
+  <p>If <a for="gb18030 encoder">is GBK</a> is false and there is a row in the table below whose
+  first column is <var>code point</var>, then return the two bytes on the same row listed in the
+  second column:
```
annevk, line 2697: As long as we layer this on top of the existing gb18030 index, we can remove all the U+EXXX PUA entries below as they are already in that table and map to the correct bytes.

An implementation that updates that table would need these U+EXXX mappings though, as evidenced by WebKit. (Which I assume has a completely separate implementation for GBK.)

With those U+EXXX mappings removed however it probably makes the most sense to create a small "gb18030 2022 index" which can be reused across the decoder and encoder as a special case when GBK is not in use. Bit more work, but overall would present the data more neatly.

annevk: WebKit (and presumably Chromium) having updated the table (as evidenced by https://github.com/vitorroriz/WebKit/commit/b7d4f07a3aac04e70f1d91d09be57e4dd2ded612) means GBK is impacted at the moment as well. In particular for the symmetrical mappings at the top of https://github.com/whatwg/encoding/issues/312#issuecomment-2354764499. What do we want to do for GBK? Perhaps it's okay if that gets slightly less PUA as well? @hsivonen what do you think?

annevk:
Closing this in favor of #336.

encoding.bs
StoneChi8: GBK-1995 never been an offical standard, although it asign some characters in U+E8xx for GB+FExx, and 52 Chinese character has been replaced by GB18030-2000 for unicode extension A, remained GB code GB+FExx unchanged, ex. GB+FE9F【䶮】 mapped to U+4DAE instead of U+E863.   For information interchange, we should use official GB18030-2022 mapping table enven in GBK quotation, in order to drop these duplicate unicode code to those same GB 2 bytes character.
Windows CP 936 method is wrong way to reach the GB18030 standard, i.e. remained U+E8xx characters in ttf font file (Source Han Sans & iOS never done these)，convert program using these PUA characters and assign 0x3F to 4 bytes GB18030 characters. 
On the other hand, full BMP PUA code range is U+E000~U+F8FF, and SMP U+10000~U+10FFFF mapping to GB18030 is a GB+91308130~ only, no mapping table need for programming and future GB18030 amendments.  
See detail in https://zhuanlan.zhihu.com/p/661610604 for WAHTWG GB18030 convert program(in Chinese).

PR #345, opened 2025-04-11
Explain the relationship between windows-1252, Latin1, and ASCII

domenic:
***
<a href="https://whatpr.org/encoding/345.html" title="Last updated on Apr 17, 2025, 7:59 AM UTC (cf74334)">Preview</a> | <a href="https://whatpr.org/encoding/345/9477a2e...cf74334.html" title="Last updated on Apr 17, 2025, 7:59 AM UTC (cf74334)">Diff</a>

### DIFF
--- a/encoding.bs
+++ b/encoding.bs
@@ -747,13 +747,13 @@ part of the ISO 8859 series. In particular, the necessity of the inclusion of <a
 and <a>ISO-8859-16</a> is doubtful for the purpose of supporting existing content, but there are no
 plans to remove these.</p>
 
-<div class=note id=note-latin1-ascii> <p>The <a>windows-1252</a> <a for=/>encoding</a> has various
-<a for=encoding>labels</a>, such as "<code>latin1</code>", "<code>iso-8859-1</code>", and
-"<code>ascii</code>", which have historically been confusing for developers. On the web, and in any
-software that seeks to be web-compatible by implementing this standard, these are synonyms:
-"<code>latin1</code>" and "<code>ascii</code>" are just labels for <a>windows-1252</a>, and any
-software following this standard will, for example, decode 0x80 as U+20AC (€) when asked for the
-"Latin1" or "ASCII" decoding of that byte.
+<div class=note id=note-latin1-ascii>
+ <p>The <a>windows-1252</a> <a for=/>encoding</a> has various <a for=encoding>labels</a>, such as
+ "<code>latin1</code>", "<code>iso-8859-1</code>", and "<code>ascii</code>", which have historically
+ been confusing for developers. On the web, and in any software that seeks to be web-compatible by
+ implementing this standard, these are synonyms: "<code>latin1</code>" and "<code>ascii</code>" are
+ just labels for <a>windows-1252</a>, and any software following this standard will, for example,
+ decode 0x80 as U+20AC (€) when asked for the "Latin1" or "ASCII" decoding of that byte.
 
  <p>Software that does not follow this standard does not always give the same answers. The root of
  this is that the original document that specified Latin1 (ISO/IEC 8859-1) did not provide any
--- a/encoding.bs
+++ b/encoding.bs
@@ -747,29 +747,28 @@ part of the ISO 8859 series. In particular, the necessity of the inclusion of <a
 and <a>ISO-8859-16</a> is doubtful for the purpose of supporting existing content, but there are no
 plans to remove these.</p>
 
-<div class=note id=note-latin1-ascii>
- <p>The <a>windows-1252</a> <a for=/>encoding</a> has various <a for=encoding>labels</a> like
- "<code>latin1</code>", "<code>iso-8859-1</code>", "<code>ascii</code>", etc. which have
- historically been confusing for developers. On the web, and in any software that seeks to be
- web-compatible by implementing the Encoding Standard, these are synonyms: "<code>latin1</code>" and
- "<code>ascii</code>" are just labels for <a>windows-1252</a>, and any software following this
- standard will, for example, decode 0x80 as U+20AC (€) when asked for the Latin1 or ASCII decoding
- of that byte.
+<div class=note id=note-latin1-ascii> <p>The <a>windows-1252</a> <a for=/>encoding</a> has various
+<a for=encoding>labels</a>, such as "<code>latin1</code>", "<code>iso-8859-1</code>", and
+"<code>ascii</code>", which have historically been confusing for developers. On the web, and in any
+software that seeks to be web-compatible by implementing this standard, these are synonyms:
+"<code>latin1</code>" and "<code>ascii</code>" are just labels for <a>windows-1252</a>, and any
+software following this standard will, for example, decode 0x80 as U+20AC (€) when asked for the
+"Latin1" or "ASCII" decoding of that byte.
 
- <p>Software that does not follow the Encoding Standard does not always give the same answers. The
- root of this is that the original document that specified Latin1 (ISO/IEC 8859-1), did not provide
- any mappings for bytes in the inclusive ranges 0x00–0x1F or 0x7F–0x9F. Similarly, the original
+ <p>Software that does not follow this standard does not always give the same answers. The root of
+ this is that the original document that specified Latin1 (ISO/IEC 8859-1) did not provide any
+ mappings for bytes in the inclusive ranges 0x00 to 0x1F or 0x7F to 0x9F. Similarly, the original
  documents that specified ASCII (ISO/IEC 646, among others) did not provide any mappings for bytes
- in the inclusive range 0x80–0xFF. This means different software has chosen different code point
+ in the inclusive range 0x80 to 0xFF. This means different software has chosen different code point
  mappings for those bytes when asked to use Latin1 or ASCII encodings. Web browsers and
  browser-compatible software have chosen to map those bytes according to <a>windows-1252</a>, which
- is a superset of both, and this was codified in the Encoding Standard. Other software throws
- errors, or uses <a>isomorphic decoding</a>, or other mappings. [[ISO8859-1]] [[ISO646]]
+ is a superset of both, and this choice was codified in this standard. Other software throws errors,
+ or uses <a>isomorphic decoding</a>, or other mappings. [[ISO8859-1]] [[ISO646]]
 
  <p>As such, implementers and developers need to be careful whenever they are using libraries which
  expose APIs in terms of "Latin1" or "ASCII". It's very possible such libraries will not give
- answers in line with the Encoding Standard, if they have chosen other behaviors for the bytes which
- were left undefined in the original specifications.
+ answers in line with this standard, if they have chosen other behaviors for the bytes which were
+ left undefined in the original specifications.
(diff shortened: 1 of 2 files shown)


encoding.bs
```diff
@@ -568,7 +580,10 @@ prescribes, as that is necessary to be compatible with deployed content.
   <tr><td>"<code>windows-1251</code>"
   <tr><td>"<code>x-cp1251</code>"
   <tr>
-   <td rowspan=17><a>windows-1252</a>
+   <td rowspan=17>
+    <a>windows-1252</a>
+    <p class=note>See <a href="#note-latin1-ascii">below</a> for the relationship to historical
+    "Latin1" and "ASCII" concepts.
```
annevk, line 586: Can you also add this to the Python script that generates this table?

encoding.bs
```diff
@@ -732,6 +747,30 @@ part of the ISO 8859 series. In particular, the necessity of the inclusion of <a
...
 plans to remove these.</p>
 
+<div class=note id=note-latin1-ascii>
+ <p>The <a>windows-1252</a> <a for=/>encoding</a> has various <a for=encoding>labels</a> like
+ "<code>latin1</code>", "<code>iso-8859-1</code>", "<code>ascii</code>", etc. which have
+ historically been confusing for developers. On the web, and in any software that seeks to be
+ web-compatible by implementing the Encoding Standard, these are synonyms: "<code>latin1</code>" and
+ "<code>ascii</code>" are just labels for <a>windows-1252</a>, and any software following this
+ standard will, for example, decode 0x80 as U+20AC (€) when asked for the Latin1 or ASCII decoding
+ of that byte.
```
annevk, line 757: I think overall this is probably okay, but what gives me pause is that the Encoding standard doesn't define Latin1 or ASCII encodings (it only defines them as labels). So if software exposes those encodings, who knows what they might do. So perhaps we should make that distinction clearer, in that this will likely happen for software that takes a label and some bytes as input.

encoding.bs
```diff
@@ -732,6 +747,30 @@ part of the ISO 8859 series. In particular, the necessity of the inclusion of <a
 and <a>ISO-8859-16</a> is doubtful for the purpose of supporting existing content, but there are no
 plans to remove these.</p>
 
+<div class=note id=note-latin1-ascii>
+ <p>The <a>windows-1252</a> <a for=/>encoding</a> has various <a for=encoding>labels</a> like
+ "<code>latin1</code>", "<code>iso-8859-1</code>", "<code>ascii</code>", etc. which have
+ historically been confusing for developers. On the web, and in any software that seeks to be
+ web-compatible by implementing the Encoding Standard, these are synonyms: "<code>latin1</code>" and
```
annevk, line 754: I vaguely recall we decided to lowercase standard? Do I misremember?

aphillips:
I added the tracker label to remind myself to come back and make a longer comment/suggestion. While Encoding really doesn't need to present the "complete and accurate history of character encodings" to explain the relationship between 1252 and L1, I do think that the text proposed here might not be clear enough/direct enough. Users of "other software" still often make use of 8859-1 as an isomorphic encoding outside of the Web context and I have encountered people who are pedantic about the Official Meaning of the 8859-1 label. I think the note being added here could be tighter and clearer about this. In particular, I find it misleading (even if factual) to say that ISO/IEC 8859-1 didn't provide mappings for the C0/C1 range and DEL. Every encoder I've ever met has mapped these isomorphically to Unicode, which puts controls at those code points.

This is the kind of thing which might merit a short external document in the I18N space so that Encoding can point (for people who want history) while focusing on "Latin1 is not an isomorphic encoding on the Web, eh?"

domenic:
Thanks @aphillips. I hope we can move forward with something basically in this form though without major objections.

> In particular, I find it misleading (even if factual) to say that ISO/IEC 8859-1 didn't provide mappings for the C0/C1 range and DEL.

I don't understand this point. I think it's pretty important to be clear about how the original standards didn't give mappings here, and I don't think it's misleading.

> Every encoder I've ever met has mapped these isomorphically to Unicode, which puts controls at those code points.

Well, except for all the ones that implement the Encoding Standard, including web browsers, right? Your statement seems pretty wrong in light of that.

encoding.bs
domenic: I tried to phrase this carefully to avoid giving the impression that latin1 or ASCII are encodings, and instead be clear that they are inputs to the common algorithm category that takes (byte sequence, encoding label) parameters.

On the web that algorithm category is well-formalized with the concepts of actual encodings vs. labels, but in larger software it's more vague with e.g. functions named `DecodeLatin1` or similar.

My attempt was "when asked for the Latin1 or ASCII decoding of that byte", but if you have a different suggestion I'd be interested. The main thing is that I don't want to only constrain us to describing the web API case where we have a clear label/encoding divide, but instead the more general category of "please decode some bytes" algorithms across all software.

encoding.bs
domenic: I think it's like the usual proper noun vs. not situation? Encoding Standard, web standards, etc.

aphillips:
> Well, except for all the ones that implement the Encoding Standard, including web browsers, right? Your statement seems pretty wrong in light of that.

You're right. I did not mean those implementing the Encoding Standard. I had in mind non-Web converters that implement Latin-1 as an actual independent encoding.

> I think it's pretty important to be clear about how the original standards didn't give mappings here

I think that's the primary thing that caused my reaction. ISO 8859 didn't specify characters in the C0 and C1 ranges, but it reserved those code points with an expectation that they'd be filled in by ISO 6429. It wasn't like escapes and controls were unknown (particularly the C0 variety). In practice, coders incorporated the controls. Unicode itself is isomorphic with 8859-1 plus 6429.

I don't think I need to repeat the history and why Encoding is based on windows-1252. What I was trying to say hurriedly before was basically: at least some developers remain familiar with non-Encoding-based coders (iconv, JDK, ICU, various databases, etc. etc.) where the label for ISO 8859-1 is an isomorphic encoding. Those developers may have used the encoding to smuggle bytes into or out of strings... and Encoding breaks that assumption (which is the whole point of your PR).

All that said... I re-read the text you have this morning, with an eye towards suggesting replacement text, and it seemed adequate this time around. So never mind. I will still raise this with I18N to decide if we might provide some exterior documentation about the history, but nothing Encoding would need to pay attention to or wait on.

encoding.bs
annevk: I found https://github.com/whatwg/html/issues/1156#issuecomment-428346283 on this, but there must be an even older discussion.

domenic: Well huh, that feels strange to my modern eyes. But happy to align. Will push a fix.

domenic: Oh. In that comment I was suggesting searching for those strings to find things to fix.

The suggested _replacement_ in that issue is `<cite>Encoding</cite>`.

We could do that (like we do `<cite>Infra</cite>`). It feels a bit strange when a spec talks about itself with a citation, but maybe it's most consistent.

Let me know what you think.

annevk: How about "this standard"?

encoding.bs
```diff
@@ -732,6 +747,30 @@ part of the ISO 8859 series. In particular, the necessity of the inclusion of <a
 and <a>ISO-8859-16</a> is doubtful for the purpose of supporting existing content, but there are no
 plans to remove these.</p>
 
+<div class=note id=note-latin1-ascii>
+ <p>The <a>windows-1252</a> <a for=/>encoding</a> has various <a for=encoding>labels</a> like
+ "<code>latin1</code>", "<code>iso-8859-1</code>", "<code>ascii</code>", etc. which have
```
annevk, line 752: like and etc. seems redundant? Maybe "labels, such as latin1, iso-..., and ascii, which have ..."?

encoding.bs
annevk: The problem I have with this is that browsers typically have "Latin1" code paths that are very much aligned with the Unicode view of the world and not windows-1252. So for complicated software it very much depends on how or what you ask.

I also don't really have a good rephrasing that would account for that. Maybe put Latin1 and ASCII in quotes like below?

encoding.bs
```diff
@@ -732,6 +747,30 @@ part of the ISO 8859 series. In particular, the necessity of the inclusion of <a
...
+ historically been confusing for developers. On the web, and in any software that seeks to be
+ web-compatible by implementing the Encoding Standard, these are synonyms: "<code>latin1</code>" and
+ "<code>ascii</code>" are just labels for <a>windows-1252</a>, and any software following this
+ standard will, for example, decode 0x80 as U+20AC (€) when asked for the Latin1 or ASCII decoding
+ of that byte.
+
+ <p>Software that does not follow the Encoding Standard does not always give the same answers. The
+ root of this is that the original document that specified Latin1 (ISO/IEC 8859-1), did not provide
+ any mappings for bytes in the inclusive ranges 0x00–0x1F or 0x7F–0x9F. Similarly, the original
```
annevk, line 761: Nit: I think whenever we talk about ranges it's always in the form of "0x7F to 0x9F".

encoding.bs
domenic: I don't think I fully understand what's making you uneasy, but I think adding the quotes is reasonable, so I'll do that.

annevk: Well, browsers typically have Latin1 or ASCII encoding implementations that don't do windows-1252. But obviously they also "Latin1" and "ascii" labels to windows-1252. So they're on both sides of the divide you're trying to draw.

tools-label-table.py
```diff
@@ -14,7 +14,14 @@ def create_table():
...
 
-            table += "  <tr>\n   <td" + rowspan + "><a>" + encoding["name"] + "</a>"
+            if encoding["name"] != "windows-1252":
+                table += "  <tr><td" + rowspan + ">" + encoding["name"] + "\n"
+            else:
+                table += f"""  <tr>
+   <td{rowspan}>
+    <a>{encoding["name"]}</a>
+    <p class=note>See <a href="#note-latin1-ascii">below</a> for the relationship to historical
+    "Latin1" and "ASCII" concepts."""
```
annevk, line 24: I suspect this is missing a newline at the end.

encoding.bs
```diff
@@ -732,6 +747,29 @@ part of the ISO 8859 series. In particular, the necessity of the inclusion of <a
 and <a>ISO-8859-16</a> is doubtful for the purpose of supporting existing content, but there are no
 plans to remove these.</p>
 
+<div class=note id=note-latin1-ascii> <p>The <a>windows-1252</a> <a for=/>encoding</a> has various
```
annevk, line 750: A newline got eaten here. And indentation is messed up as a result?

annevk (approved):
Looks good modulo formatting nits.

tools-label-table.py
domenic: I added one, and indeed the Python script output is more consistent now, but the main text doesn't match the Python script output: all the blank lines are omitted. Not sure what to do about that, if anything.

annevk: I took a more detailed look and pushed a fixup. Should be okay to land now from my perspective.

